PR #842, opened 2014-12-19
Block publishing

amberwebb:
Depends on https://github.com/Connexions/cnx-authoring/issues/173

### DIFF
--- a/src/scripts/models/content.coffee
+++ b/src/scripts/models/content.coffee
@@ -227,3 +227,11 @@ define (require) ->
         return @get('currentPage')
 
       return @
+
+    isPublishable: () ->
+      book = @isBook()
+
+      if book and @get('areContainedPublishable') or not book and @get('isPublishable')
+        return true
+      else
+        return false
--- a/src/scripts/models/content.coffee
+++ b/src/scripts/models/content.coffee
@@ -4,6 +4,7 @@ define (require) ->
   settings = require('settings')
   Collection = require('cs!models/contents/collection')
   Page = require('cs!models/contents/page')
+  $ = require('jquery')
 
   return class Content extends Collection
     relations: [{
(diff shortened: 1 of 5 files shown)


src/scripts/modules/media/editbar/block-publish/modals/block-publish-template.html
```diff
@@ -0,0 +1,18 @@
+<div id="reject-publish-modal" class="modal fade">
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <form class="form-horizontal">
+        <div class="modal-header">
+          <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>
+          <h4 class="modal-title">You Cannot Publish</h4>
+        </div>
+        <div class="modal-body">
+          You cannot publish because of the following reasons: {{publishBlockers}}
```
amberwebb: That's another way to do it. I don't know if it's necessarily a more appropriate way. A new Handlebars helper will have to be created to format the blockers, or the format helper will need to be modified to use in multiple places.

src/scripts/models/content.coffee
```diff
@@ -4,6 +4,7 @@ define (require) ->
   settings = require('settings')
   Collection = require('cs!models/contents/collection')
   Page = require('cs!models/contents/page')
+  $ = require('jquery')
```
amberwebb: Thanks for the feedback. It would probably be good to go through all the views and make sure that all dependencies are ordered that way.

src/scripts/modules/media/editbar/block-publish/modals/block-publish-template.html
dak: You don't need to write a handlebars helper for this.  The `templateHelpers` property exists to allow you to transform/append any data before passing it to the template.

Issue #808, opened 2014-12-03
Add page modal needs validation


amberwebb:
I am able to click submit on the add page modal with a blank title.

Issue #809, opened 2014-12-03
New Book or New Page modal needs cross-browser validation


amberwebb:
The HTML5 required attribute does not work in Safari. Probably need to create a validation function that can be used everywhere.

PR #844, opened 2014-12-23
Validation for add page modal

amberwebb:
Fixes https://github.com/Connexions/webview/issues/808 and https://github.com/Connexions/webview/issues/809


src/scripts/helpers/validation.coffee
```diff
@@ -0,0 +1,13 @@
+define (require) ->
+  $ = require('jquery')
+
+  return new class Validation
```
amberwebb: This would clean up the code considerably.

Issue #839, opened 2014-12-18
Successful 'Update' on role-acceptance page when nothing is selected


pumazi:
It's possible to select the 'Update' button within the role-acceptance page without having selected one or more of the roles. The message that is returned in the interface leads one to believe they have successfully accepted their role.

I'd suggest we disable the button when one or more has not been selected. The cues need to be clearer as to what can be selected in this interface.

There may be an issue in authoring, since both authors are in an unknown acceptance state. We'll need to analyze the path taken by @rpcode2014 and @nyxer95.

(CC @edwoodward @rpcode2014  Followup to Connexions/cnx-archive#276 )

amberwebb:
I think this issue is partially fixed now that publish blockers have been added.

PR #855, opened 2015-01-13
Disabling submit button unless all roles are accepted or rejected

amberwebb:
Fixes https://github.com/Connexions/webview/issues/839

### DIFF
--- a/src/scripts/modules/role-acceptances/role-acceptances.coffee
+++ b/src/scripts/modules/role-acceptances/role-acceptances.coffee
@@ -72,7 +72,7 @@ define (require) ->
     disableOrEnableSubmit: () ->
       submit = @$el.find('.submit')
       roles = @model.get('roles')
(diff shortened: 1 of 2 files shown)


src/scripts/modules/role-acceptances/role-acceptances.coffee
```diff
@@ -65,3 +68,12 @@ define (require) ->
...
 
     licenseRequired: () -> !!_.findWhere(@model.get('roles'), {hasAccepted: true})
+
+    disableOrEnableSubmit: () ->
+      submit = @$el.find('.submit')
+      roles = @model.get('roles')
+      role = _.pluck(roles, 'hasAccepted')
+      if _.contains(role, null)
+        submit.prop('disabled', true)
+      else
```
dak: Since we're already re-rendering the template on a role change, can this be handled in the template?

PR #865, opened 2015-01-16
Publish blockers error

amberwebb:


amberwebb:
Cannot publish if a page has a pending role request. Getting 400 Bad Request error in authoring.

PR #866, opened 2015-01-17
Publish blockers fix for published content

amberwebb:
This code checks to see if book is publishable, content in book is publishable and if individual pages are publishable. Tested with published pages, draft pages and books.


amberwebb:
In progress label added because of view not rendering when role removed from page.

Issue #572, opened 2014-06-17
Open editor after adding page/book to workspace


dak:
Ask for the name of the book/page first, then redirect.

Issue #872, opened 2015-01-21
Workspace with new Page displays 0 item count 


karenc:
Original ticket: https://github.com/Connexions/cnx-authoring/issues/187

Derek Ford reported

After creating a new Page it shows up in my Workspace but the workspace still says 0 items.

My comment: I think this is a bug in webview. Right after creating a new page with a title, the new page already appears in my workspace but it still says "0 items". If I refresh the page, then it says "1 items".

Issue #870, opened 2015-01-21
Book Editor - New collection should open with TOC open


edwoodward:
Users are reporting that they cannot find the Add button on a new Book.  We should open a new Book with the TOC open so the Add button is easier to find.

reedstrm:
I bet we get report a about it being hard to find the book you just created in the workspace, too. (eventually) Could kill both those by redirecting after the create-new popup title modal to the ContentView of the book. We can't just always open books w/ the TOC open, because of the "page is below the fold" problem for any book of reasonable length.

PR #878, opened 2015-01-23
navigate to newly created content

reedstrm:
this solves #872, and is a step towards solving #870


amberwebb:
Also fixes https://github.com/Connexions/webview/issues/572

Issue #885, opened 2015-01-28
Titles have underscores


edwoodward:
Any new Page or Book that is created ends up with underscores in the title once it is saved. The titles end up being similar to the titles in the URL. This started yesterday.

PR #887, opened 2015-01-28
Removing underscores in media titles

amberwebb:
Fixes https://github.com/Connexions/webview/issues/885


amberwebb:
Don't merge this. Even though this fix removes the underscores from the title, there is a bigger issue that needs to be addressed. In contents.coffee, options.title does not reference the correct title.

Issue #882, opened 2015-01-26
Browser Back Button has multiple entries for each page


edwoodward:
I think we fixed this before, but it is back.  If you click on any of the featured books and then click the back button, you cannot leave the featured book.  The are multiple entries in the history for each page.
![screenshot from 2015-01-26 14 24 09](https://cloud.githubusercontent.com/assets/1394695/5907739/155a681c-a567-11e4-9c1c-57a437d709a6.png)

amberwebb:
What is happening is that when you go into a featured book or any book for that matter, the book is being loaded twice. The first time, /contents/778e36af-4c21-4ef7-9c02-dae860eb7d14 is being loaded and pushed to history and the second time contents/778e36af-4c21-4ef7-9c02-dae860eb7d14@9.71:1/Fundamentals_of_Electrical_Eng is loaded and pushed to history.  Any time content is loaded, the version, page number and title are added to the link. This is what is causing multiple entries. When you hit the back button to get out of the book, you are navigating to the link without the version and title and then the link with the version and title is loaded, so you are stuck in this weird loop. 

Technically the back button and history are working correctly, it is the way we are loading or creating the content links that are causing the problem.

Looking for a solution to this.

edwoodward:
This is still happening on dev.cnx.org. I tested it earlier in the day and it was working correctly.  Tonight, content gets stuck on the first Page of a Book displayed. If you click into abook, it will back up until you reach the first Page and then you are stuck on that page.

PR #893, opened 2015-01-30
Removing extra content stack from history when navigating to a book

amberwebb:
Fixes https://github.com/Connexions/webview/issues/882

### DIFF
--- a/src/scripts/helpers/backbone/history.coffee
+++ b/src/scripts/helpers/backbone/history.coffee
@@ -5,7 +5,6 @@ define (require) ->
   hashStripper = /[#].*$/
   routeStripper = /^[#\/]|\s+$/g
   trailingSlash = /\/$/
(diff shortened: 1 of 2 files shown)


dak:
This seems like it's curing the symptom instead of the problem.

amberwebb:
Yes, but I am unable to find the root of the problem.

dak:
If the url duplication is happening when the url is changed to include the version and page number, then that's where the problem is.  It may just need an `options.replace = true` to make sure it replaces the url instead of adding a new one.

amberwebb:
That's actually what I've been looking for. This is happening after media view loads. What loads after media view?

amberwebb:
So I found the proper place to add options.replace. It's buried in media/tabs, which is the one place I did not look yesterday.

PR #895, opened 2015-01-30
keep the query string on rewrite of url

### LINKED ISSUE #894: TOC does not open when editing a new Book
This worked yesterday so it is a regression. A new book opens in the Editor, but does not open the TOC.

reedstrm:
Fixes #894


dak:
:+1:

PR #905, opened 2015-02-11
Remove CSS numbering from footnotes

### LINKED ISSUE #903: Double numbering footnotes
Footnotes are getting double numbered. The markup already contains a number and then CSS adds another list number. The number in the footnote is also an anchor that is both named and targeted (back to the original reference).

Example at http://dev.cnx.org/contents/fd44afbb-3167-4b83-8508-4e70885b6136@2/As_Transparent_as_Infrastructu#footnote4

dak:
Fixes #903


edwoodward:
Is this to solve the double numbering issue?

reedstrm:
yes - the description links to #903 "Double Numbering footnotes"

PR #907, opened 2015-02-11
Weekly Updates

dak:


reedstrm:
Hmm, I don't think I did anything to this PR - perhaps my playing w/ huboard against this repo did something?

dak:
@reedstrm I've noticed any issue I touch or PR I create now is automatically marked as `in progress` by you, so I assumed something you were testing with that fake issue was doing it.

PR #908, opened 2015-02-11
WIP: Upgrade to Select2 4.0

dak:
Branches off #907


reedstrm:
I'm a little concerned about doing library upgrades right now, while we're trying to stabilize the release. Are those all backwards compatible?

dak:
I'll leave this one as a WIP.  The other upgrades from #907 are all backwards compatible and mostly minor version bumps were getting from bower automatically anyway.

PR #911, opened 2015-02-12
Remove all e-mail links

### LINKED ISSUE #846: Email address missing from author email link
Click on author name in book or page. Email opens with the book title as the subject, but no email address. 'Mailto' value is missing.

dak:
Fixes #846 

Per the discussion, since e-mails will not be provided to webview in the near future and we have not yet built a system to e-mail authors indirectly yet, all e-mail links have been removed.


reedstrm:
@dak should I leave the branch extant to make it easy to find all the places to add back, when the new
system exists, or do you want to drop some sort of tag here?

edwoodward:
I just deleted it.  I missed your comment.

Issue #185, opened 2013-11-08
Select dropdowns in Firefox on Mac and Linux are ugly


dak:
See: https://github.com/twbs/bootstrap/issues/765
https://bugzilla.mozilla.org/show_bug.cgi?id=649849
https://github.com/silviomoreto/bootstrap-select
http://stackoverflow.com/questions/5912791/how-to-remove-the-arrow-from-a-select-tag-in-firefox

edwoodward:
Tested in Firefox 35.0.1 on Ubuntu and the dropdown is unchanged
![screenshot from 2015-02-25 14 47 47](https://cloud.githubusercontent.com/assets/1394695/6380224/ad46361c-bcfd-11e4-92a6-6e57211976e7.png)

amberwebb:
@edwoodward They are unchanged in Firefox on Mac too. Are we sure this change was deployed?

edwoodward:
User error. Used wrong server for test.  It is working on demo.cnx.org.  Much better.

reedstrm:
Righto. Has this been merged to master as well?

PR #920, opened 2015-02-23
Select arrows in Firefox

amberwebb:
https://bugzilla.mozilla.org/show_bug.cgi?id=649849 has now been resolved. -moz-appearance: none removes the default arrow from the select element in Firefox and I have added arrows similar to Chrome as a background image.

Fixes https://github.com/Connexions/webview/issues/185


reedstrm:
So, for this styling issue we're saying 'We support Firefox 35 and above' ? Not arguing, just clarifying.

amberwebb:
@reedstrm Good point, but I think this is the best option. The older workarounds are unreliable.

reedstrm:
works for me, as it's not a functionality diff.

Issue #853, opened 2015-01-06
Default editor text shows briefly in published page during loading


amberwebb:
http://dev.cnx.org/contents/fd53eae1-fa23-47c7-bb1b-972349835c3c@2.1:1/Precalculus

1. Click the next button.
2. "Enter some content here..." flashes while the page is loading.

PR #922, opened 2015-02-26
Default editor text showing

amberwebb:
Fixes https://github.com/Connexions/webview/issues/853


dak:
Not sure if this is a regression in this PR or currently exists, but the progress bar is not visible when a page is loading now.

amberwebb:
@dak  This seems to be an existing bug. I will create an issue.

test/scripts/models/contents/page.js
```diff
@@ -0,0 +1,32 @@
+describe('test page model', function () {
+  'use strict';
+  var Page,
+    expect = chai.expect;
+
+  beforeEach(function (done) {
+    require(['cs!models/contents/page'], function () {
+      Page = arguments[0];
```
dak: Rather than setting this to `arguments[0]`, `Page` could just be moved up into the function params.

test/scripts/models/contents/page.js
```diff
@@ -0,0 +1,32 @@
...
+    });
+  });
+  describe('results returned', function () {
+    it('should show default text if page is loaded and content is empty', function () {
+      var page = new Page();
+      var results = {
+        content: '',
+        loaded: true
+      };
+      if (results.loaded && results.content === '') {
```
dak: Since `results.loaded` is hardcoded to be `true`, is it even necessary to set for it (or set it?).

test/scripts/models/contents/page.js
```diff
@@ -0,0 +1,32 @@
...
+    });
+  });
+  describe('model defaults', function () {
+    it('should have default type for mediaType', function () {
+      var page = new Page();
+      expect(page.get('mediaType')).to.equal('application/vnd.org.cnx.module');
+    });
+  });
+  describe('results returned', function () {
+    it('should show default text if page is loaded and content is empty', function () {
```
dak: This doesn't appear to test for default text, only that a page's content is what it is set to be.

dak:
Do we want to be writing tests using `Should`, `Expect`, or `Assert`?  I believe current tests are written using the `should` style.

dak:
Did you want to squash these down to 1 commit?

src/scripts/modules/workspace/popovers/new/modals/new-media.coffee
```diff
@@ -26,6 +26,7 @@ define (require) ->
       else
         content = new Page
           title: options.title
+          content: '<p>Enter some content here. Format content, and drag and drop elements from the toolbar.</p>'
```
dak: Is it necessary to set the content here as well?  Should it not be visible on the page when it loads and has no content?

amberwebb: The stand alone page does not show the default text when the content is set in toJSON method.

dak: Hmm, this makes me think we may have more hidden regressions because of this.  Some places may be calling `toJSON` and others may be getting it directly with `get('contents')`.

The default text should only be loading and visible after the page has loaded and been overwritten by the AJAX response, so what we're doing here may be more of a hack for the symptom instead of the problem.

I think this may be closely related to the issue of the progress bar not showing up when loading a page either.  There's probably a logic error in the body template to decide whether or not to show the content.  The content shouldn't be shown at all until the page is loaded, whereas right now it's showing the content (whether the default text or empty content).

amberwebb: I'm going to pull down this PR and investigate the issue more closely.

PR #929, opened 2015-03-04
Show progress bar when page is loading

### LINKED ISSUE #853: Default editor text shows briefly in published page during loading

dak:
Fixes #853

### DIFF
--- a/src/scripts/modules/media/body/body-template.html
+++ b/src/scripts/modules/media/body/body-template.html
@@ -1,12 +1,6 @@
-{{#is loaded false}}
-  <div class="media-body">
-    <div class="progress progress-striped active">
-      <div class="progress-bar secondary" role="progressbar" style="width: 100%">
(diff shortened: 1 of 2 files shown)


dak:
Per the discussion realization in #922, this PR fixes the logic to hide page content when a page is still loading.

This appears to work well on a normal page; however, I do believe the logic was previously changed due to an issue related to how pages were displayed during editing, so this is in need of some extensive testing to ensure it doesn't just trade one bug for another.

amberwebb:
Good call on this. This fixes the issue with the progress bar not loading and the default content only showing after a page is loaded, although now the progress bar is showing after creating a book and before creating a page in that book.

PR #940, opened 2015-03-05
Make devarchive the default cnx-archive server

dak:


amberwebb:
I guess settings.js doesn't get deployed to demo and beta?

dak:
`settings.js` is used for all deployments, but @reedstrm maintains those for each one.  The default is really just used for development.

PR #951, opened 2015-03-10
Feature/embed external exercises

pandafulmanda:
Unsure if i should be doing a PR?  Seems like a good idea.

will create separate branch for body.coffee renderDom refactor

### DIFF
--- a/src/data/exercises.json
+++ b/src/data/exercises.json
@@ -1,0 +1,455 @@
+{
+    "total_count": 10,
+    "items": [
+        {
+            "uid": "k12phys-ch04-ex034",
+            "number": 1,
+            "version": 1,
+            "license": {
+                "name": "cc_by_4_0",
+                "title": "Creative Commons Attribution 4.0 International",
+                "url": "http://creativecommons.org/licenses/by/4.0/",
+                "is_attribution_required": true,
+                "is_share_alike_required": false,
+                "is_derivation_allowed": true,
+                "is_commercial_use_allowed": true
+            },
+            "editors": [],
+            "authors": [],
+            "copyright_holders": [],
+            "derived_from": [],
+            "attachments": [],
+            "stimulus_html": "",
+            "questions": [
+                {
+                    "stimulus_html": "",
+                    "stem_html": "<p>A spherical conductor has a net positive charge. Describe how the charge is distributed around the sphere's surface.</p>",
+                    "answers": [
+                        {
+                            "content_html": "<p>All positive charge is on the sphere surface, and all negative charge is at the center.</p>"
+                        },
+                        {
+                            "content_html": "<p>Positive and negative charge is spread equally around the sphere's surface.</p>"
+                        },
+                        {
+                            "content_html": "<p>The net positive charge is spread evenly throughout the sphere's volume.</p>"
+                        },
+                        {
+                            "content_html": "<p>The net positive charge is spread evenly around the sphere's surface.</p>"
+                        }
+                    ],
+                    "hints": [],
+                    "formats": [
+                        "multiple-choice"
+                    ],
+                    "combo_choices": []
+                }
+            ]
+        },
+        {
+            "uid": "k12phys-ch12-ex099",
+            "number": 2,
+            "version": 1,
+            "license": {
+                "name": "cc_by_4_0",
+                "title": "Creative Commons Attribution 4.0 International",
+                "url": "http://creativecommons.org/licenses/by/4.0/",
+                "is_attribution_required": true,
+                "is_share_alike_required": false,
+                "is_derivation_allowed": true,
+                "is_commercial_use_allowed": true
+            },
+            "editors": [],
+            "authors": [],
+            "copyright_holders": [],
+            "derived_from": [],
+            "attachments": [],
+            "stimulus_html": "",
+            "questions": [
+                {
+                    "stimulus_html": "",
+                    "stem_html": "<p>Which of the following is not a property of a conductor in electrostatic equilibrium?</p>",
+                    "answers": [
+                        {
+                            "content_html": "<p>The force on a charge near the conductor is zero.</p>"
+                        },
+                        {
+                            "content_html": "<p>Electric field lines just outside the conductor are perpendicular to the surface.</p>"
+                        },
+                        {
+                            "content_html": "<p>All net charge in the conductor is on the surface.</p>"
+                        },
+                        {
+                            "content_html": "<p>The electric field inside the conductor is zero.</p>"
+                        }
+                    ],
+                    "hints": [],
+                    "formats": [
+                        "multiple-choice"
+                    ],
+                    "combo_choices": []
+                }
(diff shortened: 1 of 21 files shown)


src/data/exercises.json
```diff
@@ -0,0 +1,455 @@
+{
+    "total_count": 10,
```
philschatz: I believe this JSON is for a set of exercises but the `<a href="...">` in the content HTML points to a single exercise. Thoughts @Dantemss?

Dantemss: Correct, this is the result set with 10 exercises.

For an example of what you will get for searching for 1 embed tag, see here: https://exercises-dev1.openstax.org/api/exercises?q=tag:k12phys-ch04-ex001

Searching for multiple tags at once:
https://exercises-dev1.openstax.org/api/exercises?q=tag:k12phys-ch04-ex001,k12phys-ch04-ex002

And if you search for a tag that's in a bunch of exercises, like a LO, you will get a lot of results: https://exercises-dev1.openstax.org/api/exercises?q=tag:k12phys-ch04-s01-lo01

Dantemss: These URL's might need to change slightly once exercises is public (we don't want exercises from random people, only trusted authors or exercises in certain lists), but for now they should work :)

pandafulmanda: The JSON is a stub for the real thing that would return only the one 1 embed tag that is relevant.  I could put extra logic in to sort the "results" for items that match only the tag, but that would be unnecessary when using the real API.

I guess I put something in to handle both cases

src/scripts/configs/embeddables.coffee
```diff
@@ -0,0 +1,41 @@
+define (require) ->
+
+  _ = require('underscore')
```
dak: It doesn't appear that underscore is a dependency of this file (the only place I see it being used is commented out).

src/scripts/configs/embeddables.coffee
```diff
@@ -0,0 +1,41 @@
...
+    },{
+      match : '#ost\/api\/ex\/'
+      matchType : 'a'
+      embeddableType : 'exercise'
+      apiUrl : () ->
+        # # stub.  Comment out to use local exercises stub.
+        # # Copied from https://exercises-dev1.openstax.org/api/exercises?q=tag:k12phys-ch04-s01-lo01
+        # 'http://localhost:8000/data/exercises.json';
+
+        'https://exercises-dev1.openstax.org/api/exercises?q=tag:' + @.itemCode;
```
dak: Should this be configured in `settings.js`?

src/scripts/configs/embeddables.coffee
```diff
@@ -0,0 +1,41 @@
...
+      #     _.indexOf(item.tags, @itemCode) > -1
+      #   , @)
+
+      onRender : ($parent) ->
+        # For the maths.  you know. it do what it do.
+        $mathElements = $parent.find('[data-math]')
+        $mathElements.each((iter, element) ->
+          mathTex = '[TEX_START]' + $(element).data('math') + '[TEX_END]'
+          $(element).text(mathTex)
+        )
```
dak: We've been avoiding wrapping multi-line params in parens as a general style.  See below:

```coffeescript
$mathElements.each (iter, element) ->
  mathTex = '[TEX_START]' + $(element).data('math') + '[TEX_END]'
  $(element).text(mathTex)
```

src/scripts/modules/media/body/body.coffee
```diff
@@ -1,11 +1,24 @@
...
   Mathjax = require('mathjax')
   router = require('cs!router')
   EditableView = require('cs!helpers/backbone/views/editable')
   template = require('hbs!./body-template')
   require('less!./body')
 
+  embeddableTemplates = {
+    'exercise' : require('hbs!./embeddables/exercise-template'),
+    'iframe' : require('hbs!./embeddables/iframe-template')
+  }
```
dak: As a general style we've been dropping multi-line brackets and commas when possible:
```coffeescript
embeddableTemplates =
  'exercise' : require('hbs!./embeddables/exercise-template')
  'iframe' : require('hbs!./embeddables/iframe-template')
```

src/scripts/modules/media/body/body.coffee
```diff
@@ -125,12 +134,201 @@ define (require) ->
...
+      # finds fresh element in @$el if a selector is provided
+      # instead of the element itself.
+
+      embeddableItem.$el = @$el.find(embeddableItem.selector) if embeddableItem.selector
+
+      $parent = embeddableItem.$el.parent()
+      embeddableItem.$el.replaceWith(embeddableItem.html)
+
+      if _.isFunction(embeddableItem.onRender)
+        embeddableItem.onRender($parent)
```
dak: Coffeescript offers an easier way to check if something is a function:
```coffeescript
embeddableItem.onRender?($parent)
```

Compiles to:
```
if (typeof embeddableItem.onRender === "function") {
  embeddableItem.onRender($parent);
}
```

src/scripts/modules/media/body/body.coffee
```diff
@@ -125,12 +134,201 @@ define (require) ->
...
+    # Could abstract a little more and pull out to another file if needed.
+    ###
+
+    # A queue for rendering embeddables means
+    # we can share rendering functions regardless
+    # of whether the embeddable is sync or async
+    initializeEmbeddableQueues: () ->
+      @renderEmbeddableQueue = []
+
+      _.extend(@renderEmbeddableQueue, Backbone.Events)
```
dak: I haven't looked at this closely to comment on it intelligently, but as a quick glance this strikes me as something that needs a closer look as possibly being overkill.

dak:
@pandafulmanda It looks like the build is failing because `settings.js` is not beautified (some formatting issue probably).  Also, can you try squashing some of these commits?

src/data/exercises.json
```diff
@@ -0,0 +1,1040 @@
...
+          "formats": [
+            "multiple-choice",
+            "free-response"
+          ],
+          "combo_choices": []
+        }
+      ]
+    }
+  ]
+}
```
dak: Can you run this through `http://jsbeautifier.org` (with 2 spaces option)?

Also, for largely [historical reasons](http://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-file), all files in webview should end in a new-line (ie, have a blank last line)

src/scripts/configs/embeddables.coffee
```diff
@@ -0,0 +1,44 @@
+define (require) ->
+
+  settings = require('settings')
+
+  return {
+    embeddableTypes : [{
+      match : '#terp-'
+      matchType : 'a'
+      embeddableType : 'terp'
+      apiUrl : () ->
```
dak: There's a bunch of files in this commit with a space between the property and colon.  You may want to do a search replace for ` : ` -> `: `.

Issue #950, opened 2015-03-10
Derived Module - Have to press Save button twice before Publish button shows up


amberwebb:
Makes me think it's rendering twice somewhere.

Issue #947, opened 2015-03-09
Cannot scroll to top of page


edwoodward:
I opened a Page to edit on dev.cnx.org and I cannot scroll to the top of the page. The links to Workspace, etc. are hidden behind the editor toolbar.

amberwebb:
This issue is related to this commit: https://github.com/Connexions/webview/commit/b164af7b6ef25669805833d043b6ae6073b32601

PR #952, opened 2015-03-10
Fix for scrolling in modules

amberwebb:
Instead of proxying change event for title, the code is listening for change:changed so the view is rendered when a title or content is changed within a section.

Fixes https://github.com/Connexions/webview/issues/950
Fixes https://github.com/Connexions/webview/issues/947


dak:
@amberwebb Can you rebase this?

PR #960, opened 2015-03-13
Bug/empty search on add modal/issue 731

pandafulmanda:


amberwebb:
Error messages should be in red not brown. The error message text needs to be left justified underneath the input field.
image

The error message does not go away when I close the window.

amberwebb:
If I search for a page and then delete the search term, the validation is triggered.

image

amberwebb:
In add-page-template, instead of <p class="help-block hide"> maybe it should be <div class="alert alert-danger help-block hide"> so we can use bootstrap's alert styling. Role acceptances uses this and it looks nice. Thoughts on this?

image


The red border around the search button, the input field and the red color of the text don't look like they match each other.

pandafulmanda:
okay.

Issue #730, opened 2014-10-31
Progress bar is active in Add Page(s) modal when canceling a search


amberwebb:
In the editor, add a page. Type something in the input field, click search and then cancel. Select 'Add Page' again. The progress bar is still active.

When search is finished, it prints out the search results even though the search was cancelled.

PR #962, opened 2015-03-13
able to cancel search on modal close and clcking cancel using the query'...

pandafulmanda:
...s abort function and then emptying the results region view, #730

### DIFF
--- a/src/scripts/models/search-results.coffee
+++ b/src/scripts/models/search-results.coffee
@@ -47,7 +47,8 @@ define (require) ->
 
       if query isnt @query or url isnt @searchUrl
         @config(options)
-        @fetch(options)
+        promise = @fetch(options)
+        @set('promise', promise, {silent: true})
 
       return @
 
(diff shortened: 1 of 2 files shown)


amberwebb:
The search results are still coming up in the modal if I cancel after a successful search.

pandafulmanda:
Okay, I guess I misunderstood this to only need to clear for when the search timeouts or when you cancel in the middle of the search.

So when someone hits cancel and there are search results, should it clear the results and the search text as well?

When someone hits the black modal overlay and that closes the modal, should that cause all the same actions as when you hit "Cancel" to close the modal?

Thanks!

amberwebb:
The modal should clear when you cancel in the middle of a search. When  you reopen the modal, the progress bar should not show (which you have now) and the input field should be empty. If the modal is closed or canceled after a successful search, the results and input field should be empty as well. I don't think clicking on the overlay should close the modal, but that may be intended behavior with bootstrap modals? Not really sure about that, but it does happen with our other modals.

pandafulmanda:
That is the default behavior with bootstrap modal.  I can change that setting for this modal, or do you think all modals on the webview should behave as such?

amberwebb:
I don't know if they should or not. Make a separate issue for that and mark it as a question so we can address it as a team at a later time.

PR #986, opened 2015-03-23
update bower.json aloha hash

reedstrm:


reedstrm:
nevermind, Amber beat me to it.

PR #987, opened 2015-03-23
Remove modal backdrop on delete

amberwebb:


reedstrm:
This is fixing an bug, I assume? Is there an issue?

amberwebb:
No issue, a bug I found when testing dev.

PR #1007, opened 2015-03-30
Role acceptance link only displayed when needed

edwoodward:
Small change that fixes status link. It was pointing to role acceptance for every state except Draft.  It is now pointing to the content for every state except "Awaiting acceptance'.


dak:
:+1:

Issue #1015, opened 2015-04-01
print style values wrong


reedstrm:
Print styles are currently selected from a hardcoded list in metadata.coffee, using the human selection string 'CCAP Physics', 'CCAP Chemistry', etc. The rest of the system (and in particular, the print system) needs ccap-physics, ccap-chemistry, etc.

PR #1017, opened 2015-04-02
Fix values for print styles

amberwebb:
Fixes https://github.com/Connexions/webview/issues/1015


src/scripts/modules/media/tabs/metadata/metadata.coffee
```diff
@@ -22,19 +22,41 @@ define (require) ->
...
-        'CCAP Precalculus'
-        'CCAP Psychology'
-        'CCAP Economics'
-        'CCAP History'
-        'CCAP Chemistry'
-      ]
+      model.printStyles =
+        [{
+          'name' : 'Latex',
+          'value' : 'latex'
```
reedstrm: We need this value to be null, actually

PR #1020, opened 2015-04-02
switch to dict and bind a human name for non-edit metadata tab

reedstrm:


dak:
Build is failing because of a minor format issue:
src/scripts/modules/media/tabs/metadata/metadata.coffee
  ✖  line 27  Line ends with trailing whitespace

reedstrm:
build succeeds. Only issue seems to be that for Default it sends the literal string 'null' Seems no simple way around that in the frontend. We can fix that up in authoring

PR #1047, opened 2015-04-21
Fix for mapping priorities and styles

amberwebb:


reedstrm:
Clearly correct, not going to wait on travis, which seems backed up.

PR #1055, opened 2015-04-27
Tutor sim overlay

amberwebb:


src/scripts/modules/media/body/body.coffee
```diff
@@ -409,3 +417,10 @@ define (require) ->
 
     resetKeySequence: (e) ->
       key[e.keyCode] = false
+
+    simLink: (evt) ->
+      evt.preventDefault()
+      link = $(evt.currentTarget)
+      @model.set('simUrl', link.attr('href'))
```
philschatz: Oh, I thought the sims would still be an `<iframe>`. When did `ost-iframe-embeddable` get defined? I don't see it defined anywhere in https://github.com/openstax/tagging-legend/blob/master/README.md

edwoodward: This design was approved by Kathi in the Pivotal card.

src/scripts/modules/media/body/embeddables/modals/sims/sims-template.html
```diff
@@ -0,0 +1,13 @@
+<div id="sims-modal" class="modal fade">
+  <div class="modal-dialog">
+    <div class="modal-content">
+        <div class="modal-header">
+          <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>
+          <h4 class="modal-title">{{simTitle}}</h4>
+        </div>
+        <div class="modal-body">
+            <iframe src="{{simUrl}}" width="960px" height="785px" frameborder='0'></iframe>
```
philschatz: Oh, do all sims have the same height? I thought Patrick had a fixed width but the heights changed depending on the sim.

amberwebb:
@philschatz The sims have different heights, but a max height of 720px plus the footer. I cannot get the height of the sims through the iframe because of same-origin policy, so we decided to make all the iframes the same height within the overlays.

src/scripts/modules/media/body/embeddables/modals/sims/sims-template.html
philschatz: And if they are the same size should this be in CSS rather than HTML attributes?

PR #1058, opened 2015-04-30
Maintenance page updates

amberwebb:
I'm getting a 403 forbidden error when I reference the logo file, even though it is in the same directory. Is this an nginx permissions issue?


pumazi:
:+1:

PR #1068, opened 2015-05-11
Change ost-iframe-embeddable to os-simulation

amberwebb:


Dantemss:
You ok with this tag @philschatz? Can this be merged?

philschatz:
Checking if the additional class is necessary.

philschatz:
we use `os-embed` in other places. How about using `.ost-interactive .os-embed` as the `document.querySelector(...)` rather than defining a new class? That would match the videos a little closer as well

Dantemss:
Should `ost-interactive` become `os-interactive` then?

philschatz:
technically, probably yes. how much work would it be amanda? and the tagging-legend would need to be updated (kim/Alina)

Dantemss:
According to Amber, the link should now have `.os-interactive .os-embed`.
Tutor still needs a class on the note and on the figure. My guess is that the note is `ost-interactive` and the figure is `.os-simfig` or something?

philschatz:
Oh, I wonder if this opens up the `ost-reading-discard` discussion since the semantics is that Tutor should discard the element (but maybe look inside it for the `os-embed` element)

amberwebb:
@philschatz Is this PR okay to merge with class names .os-interactive .os-embed?

philschatz:
Yep, I think so :grimacing:. The rest is tutor-specific.

Dantemss:
So you're ok with having both os-interactive and ost-interactive tags right Phil?

PR #1074, opened 2015-05-20
Added summary to description meta tag for SEO

cburk:


src/scripts/helpers/backbone/views/base.coffee
```diff
@@ -117,7 +117,8 @@ define (require) ->
 
       if description isnt undefined
         $('meta[name="description"]').remove()
-        head.append("<meta name=\"description\" content=\"#{description}\">")
+        fullDescription = description + if summary isnt undefined then " " + summary else ""
```
dak: You may want to check for truthiness on summary rather than undefined, so that it ignores an empty string, null, and undefined.  Something like: `fullDescription = description + if summary then " #{summary}" else ''`

test/scripts/helpers/backbone/views/base.js
```diff
@@ -179,7 +186,7 @@ describe('backbone base view helper tests', function () {
         'images/social/logo.png');
       document.querySelectorAll('meta[property=\'og:image\']').length.should.equal(1);
       document.querySelector('meta[name=\'description\']').getAttribute('content').should.equal('second ' +
-        'description');
+        'description' + ' ' + 'second summary');
```
dak: You might as well turn `+ ' ' + 'second summary'` into `+ ' second summary'`.

amberwebb: @edwoodward This is a change from October. I have verified it did not get moved over to production branch.

PR #1076, opened 2015-05-21
Production Meta Tag Addition

cburk:
Added Amber's description and summary meta tag code from master


amberwebb:
I'm not exactly sure how these different commits got into your branch, but this can't be merged as is. What you can do is create a branch off of production branch and then move the code from this commit https://github.com/Connexions/webview/commit/d6e74bd9c8dd35bb91128178a6aedda1e9a21973 into your branch. You should only be touching 8 files that way.

PR #1092, opened 2015-06-04
Replaced book title with page title in URL

cburk:
And adjusted existing tests to match new layout


amberwebb:
The book title is still in the URL when the page loads, but is being redirected and then displaying the page title. There are probably other places in the code that you will need to account for. Single pages will not load. There is an error in the console to help you remediate this bug. History is not showing the page title. This needs to be addressed too.

PR #1093, opened 2015-06-05
NC-SA license change

amberwebb:


src/scripts/models/content.coffee
```diff
@@ -266,3 +266,13 @@ define (require) ->
...
+
+    license: () ->
+      license = undefined
+      @setLicense = (value) ->
+        license = value
+        return
+
+      @getLicense = ->
+        license
+      return
```
dak: I don't think you want to be returning nothing here in a 'get' method.  Since CoffeeScript implicitly returns the value of the last line, you can just write this as `@getLicense = -> license`.

Although should the `license` value actually be stored on the content model itself, rather than just existing in the license method closure?

src/scripts/modules/media/media.coffee
```diff
@@ -32,6 +33,11 @@ define (require) ->
     summary:() -> @updateSummary()
     description: () -> @updateSummary()
 
+    events: {
+      'keydown .media-title > .title input': 'checkKeySequence'
+      'keyup .media-title > .title input': 'resetKeySequence'
+    }
```
dak: Just a style comment: I believe most of the CoffeeScript code has been leaving off brackets for objects when it would be unambiguous to do so.

PR #1100, opened 2015-06-10
Rearranged title (book-page -> page-book)

cburk:


cburk:
sure, I would have used 'updateTitle' like you originally suggested, but that already existed in media.coffee

PR #1101, opened 2015-06-10
Added page titles to url

cburk:


amberwebb:
Experiencing some erratic behavior while navigating through the TOC. Some of the URLs do not have the page number and sometimes the URL doesn't show the correct title.

cburk:
Almost there I think, I just need to fix an error I noticed where the title has an unnecessary '-' appended to it

amberwebb:
@cburk I am not seeing the page showing twice in history, but if I click on a page in history, it does not take me to the correct page. In Anatomy & Physiology for example, I am taken to Preface.

amberwebb:
This looks good, only one comment on your last update. Also, when I navigate to a book from the search results, the book title flashes in the url until the page is loaded. Could this possibly be because the links in the search results have a title? src/scripts/modules/search/results/list/item-partial.html

PR #1106, opened 2015-06-24
Fixed off by 1 error in browser history

cburk:
Previously, a view could have a title that aligned with the next page to be viewed, which would create a mismatch in browser history.  This code addresses that.


amberwebb:
@cburk This branch should be a branch off url-title2 since both branches have related code. I cannot test this for merge.

PR #1120, opened 2015-06-30
Add role dialog to modals

amberwebb:


dak:
Do we have any examples of people relying on a screen reader that have run into this issue?  [The spec](http://www.w3.org/TR/wai-aria/states_and_properties#aria-hidden) does state that screen readers that `Authors are reminded that visibility:hidden and display:none apply to all CSS media types; therefore, use of either will hide the content from assistive technologies that access the DOM through a rendering engine.`

Since we're a single page app, I don't think a screen reader would be able to load or access our site if it didn't support DOM rendering (Note: It might be able to use with `_escaped_fragment_=`, but it wouldn't be very useful).  If that's the case, then we probably don't need to rely on `aria-hidden`.

dak:
Per the discussion in the Bootstrap github issue on this, it appears that the above advice is outdated and it's no longer recommended to use aria-hidden if an element's display is hidden with `display: none` or `visibility: hidden`.

http://www.w3.org/TR/wai-aria-1.1/#aria-hidden

amberwebb:
We had an independent company perform an accessibility audit on cnx.org in March 2015. The report suggested adding `aria-hidden` to visually hidden elements. The report can be found in Trello.

https://trello.com/c/cTAPWZOI/1257-add-aria-hidden-to-hidden-modals-or-text

dak:
Perhaps the company that audited is slightly out of date?  The W3C discussion linked above came to a consensus that the text was no longer correct to use (and actually causes some issues) only recently (in December 2014), and it's currently only updated in the Draft spec (although there appeared to be broad consensus for this specific change in the discussion).

PR #1122, opened 2015-07-06
Fix figure links

amberwebb:
This fixes issue where links to figure elements in content was being displayed as '[link]'. Also takes into account exercises, problems and notes where the tagName is 'div' and we need the data-attribute to display the proper link text.


philschatz:
Does this work for example for links to a figure in another module?

reedstrm:
no @philschatz we're not fetching every link to get it's tag property - this operates in the local DOM only.

philschatz:
@kjdav mentioned something similar this morning in the current cnx sprint. Is this the extent of the fix?

edwoodward:
This is all we have planned for this issue at this time.

PR #1145, opened 2015-08-04
Changed header to include rel=next/prev tag

cburk:
For SEO, google wants paginated content to define a next/prev relationship between pages.  This code makes that change by modifying the header links to include rel="next" and rel="prev" for paginated content.


dak:
@cburk I've looked at this a couple times now, and I suspect I'm probably just misunderstanding how we're using canonical links, but based on your description do we not probably want both canonical (for non-paginated content) and next/prev?

I was under the impression that we used canonical links to ensure that various different URLs were all recognized as being the same content (ie, /uuid vs. uuid@version vs. uuid@version/title).  We can get to content without using a next/previous relationship, such as when going to a book from the home page or through search, and it seems like in those cases it's necessary to let a crawler know what the canonical URL for the content it just loaded should be.

The next/prev rels would then also be relevant to place the book's page in context, but I don't think they could stand alone, unless I'm misunderstanding something.

And on a slightly separate note, if next/prev is only relevant to the media content, then it might be better to move those methods out of the base view and put them closer to the views that will actually be using them.

edwoodward:
@dak Christian is basing his work on this document: https://docs.google.com/document/d/1vGjBlz65lA3B7BTTlSECK-8W2ZpJiOmW50TXZlJPXMU/edit#

I got the information from http://googlewebmastercentral.blogspot.com/2013/04/5-common-mistakes-with-relcanonical.html

relcanonical may be ok as we are using it, but we need to add next and prev

amberwebb:
@dak Is this okay to merge?

amberwebb:
@cburk Please also make a PR for production.

cburk:
@dak, at your convenience, would you verify that the most recent commit performs the desired behavior?  It should now include the rel=canonical tag as well as the next/prev ones.  Please let me know if you still have any concerns.

dak:
@cburk Is it possible to move the next/prev methods out of the Base View and put them closer to where they're actually used (whether putting them on the media view directly or making an inherit for media to use).

It seems unnecessary to have those on the base view since they're only used by the media views, and would also allow removing the extra properties from the other pages.

cburk:
I think that's doable.  I'll be sure to look into it.

cburk:
@dak would something like this be acceptable?

dak:
:+1:

cburk:
Cool, I'll rewrite the test.  Thanks for the feedback

cburk:
@dak thanks for the suggestion about rebasing.  The new error:

(Uncaught TypeError: 'undefined' is not a function (evaluating '_math.bind(null, Math[f], mathFunctions[f])') (file:///home/vagrant/webview/bower_components/require-less/lessc.js:2801)) 

seems to be the result of line 10 in nav.coffee, which just requires the less script.  Do you know the problem here/can you suggest a workaround?

dak:
@cburk Mocha is using PhantomJS to run the tests, which uses an ancient version of Chrome that doesn't support `bind`.  You can try adding a (polyfill for `bind`)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind#Polyfill] at the top of the test.

cburk:
@dak Thanks!  I was running into trouble trying to include the bind code in nav.coffee, so I put it in a .js file in the same directory.  Does this code belong somewhere else, like in helpers?

dak:
Try this:
```cs
if (not Function.prototype.bind)
  Function.prototype.bind = (oThis) ->
    if (typeof @ isnt 'function')
      throw new TypeError('Function.prototype.bind - what is trying to be bound is not callable')

    aArgs = Array.prototype.slice.call(arguments, 1)
    fToBind = @
    FNOP = () ->
    fBound = () ->
      fToBind.apply((if this instanceof FNOP then @ else oThis),
        aArgs.concat(Array.prototype.slice.call(arguments)))

    if (this.prototype)
      FNOP.prototype = this.prototype;
    fBound.prototype = new FNOP();

    return fBound;
```

You can just add this to the top of the test that requires it.

PR #1149, opened 2015-08-10
Enhance description meta tag

cburk:
The description meta tag is currently just the book summary.  With this change, when individual pages have their own abstract/summary, the description tag is set to this instead.  Otherwise, it remains the same as before.

### DIFF
--- a/src/scripts/helpers/backbone/views/base.coffee
+++ b/src/scripts/helpers/backbone/views/base.coffee
@@ -107,7 +107,10 @@ define (require) ->
 
     addMetaTags: () ->
       summary = @summary?() or @summary
-      description = @description?() or @description
+      if typeof @description is 'function'
+        description = @description() or ''
+      else
+        description = @description
       location.origin = linksHelper.locationOrigin()
       head = $('head')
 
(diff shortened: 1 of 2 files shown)


amberwebb:
@cburk Please make a PR for production.

cburk:
Hey @amberwebb, I took a look at your comments and retested the code w/ and w/o your suggestions.  I don’t fully understand the difference between my version and the old one, but with mine the description tag in the following page displays correctly (http://10.11.12.13:8000/contents/fa91e794-a2a8-40d3-babf-e32a69e3d507@1.1:1/MY-FAMILY) whereas w/ the old code the description and og:description metatags appear as text versions of function calls.  I think the important difference is in the new line 111 of base.coffee, which works in conjunction w/ line 130 to make sure description isn’t overwritten by a function that isn’t working properly.  If you recognize this kind of error I'd appreciate any advice you have in solving it more elegantly.  As you can see, in either version the og:description tag in the above link appears as this weird string version of the function.

cburk:
Retesting this yesterday the code seemed to work.  I didn’t notice description occurring twice, is it possible that what you saw was the difference between the og:description metatag and the description metatag?

PR #1164, opened 2015-08-31
Prevent overriding styles in dist build

dak:
The `.social-button` mixin defines defaults for `background-size` and `background-position`, which should be getting overridden by the new styles defined after them (and is appropriately overridden in dev).  However, in the dist version, the CSS compressor makes an unfortunate choice that causes the rules to appear in the wrong order.  Adding `!important` ensures they will still take precedence.

### DIFF
--- a/src/styles/main.less
+++ b/src/styles/main.less
@@ -146,7 +146,7 @@ h5 {
 
 .facebook {
   .social-button("/images/social/facebook.png");
-  background-size: 0.9rem 2rem;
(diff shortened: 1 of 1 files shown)


dak:
@amberwebb It's been a while since that code was written, but I suspect it's because the LinkedIn logo is being centered and constrained, while the google logo is just being shifted.  It would probably be possible to make them use the same units, but I'm not sure it's worthwhile to do so.

PR #1169, opened 2015-09-02
New license

RoyEJohnson:
Removed empty divs and unused css, changed css class names to use
hyphenated-name instead of camelCase convention


dak:
The whitespace for the license-template is a bit awkward.  The opening `<dt>` tags are at the very end of some long lines, while their closing `</dt>` tag has no indentation at all.

PR #1172, opened 2015-09-03
Updated license formatting

RoyEJohnson:
Using h3 and p tags instead of dt and dd tags. Extraneous CSS removed.


dak:
@edwoodward @reedstrm Should we also update the license link to point to CC BY 4.0 instead of 3.0?

PR #1173, opened 2015-09-08
Upgrade to new container-based travis builds

dak:


dak:
Added to #1166 instead since waiting was getting old.

PR #1177, opened 2015-09-09
Hide carousel

RoyEJohnson:
Removed carousel code for Featured Books. Less now shows two of each group; More shows all.


amberwebb:
Shouldn't all of the books show instead of just two CNX and two OpenStax College? If I toggle the more link, the CNX books disappear.

edwoodward:
The listing should show as if the More link was selected. This means all 
the books display all of the time.

RoyEJohnson:
What I did was have it display two of each group of books, and if you click More, it shows all. That removed the carousel problem and didn't have all the books taking up screen space initially. Do we want to remove the More/Less functionality entirely and show all the books all the time?

The CNX books disappearing on going to Less is a bug on my part.

amberwebb:
@RoyEJohnson Was leaving the more link a part of Jason's design?

RoyEJohnson:
@amberwebb No, my mistake.

dak:
Rather than changing the InfoBlock inherit, it would probably make more sense to just not inherit it and change the inherit back to BaseView.  If nothing else is using the InfoBlock it could be deleted altogether.

PR #1179, opened 2015-09-11
Section modal needs styling

RoyEJohnson:
Functionally complete, but widget styling and code organization probably need to change.


src/scripts/modules/media/tabs/contents/toc/modals/section-name/section-name.coffee
```diff
@@ -0,0 +1,29 @@
...
+
+  return class SectionNameModal extends BaseView
+    template: template
+
+    events:
+      'click .btn-ok': 'ok'
+      'keydown': 'keyAction'
+
+    keyAction: (event) ->
+      if event.keyCode == 13
```
dak: `if event.keyCode is 13` (which in JavaScript will translate to `===`... never ever use `==` in JS/CS unless you really really really mean it, and even then, maybe don't).

src/scripts/modules/media/tabs/contents/toc/modals/section-name/section-name.coffee
```diff
@@ -0,0 +1,29 @@
...
+        event.stopPropagation()
+        @$el.find('.btn-ok').click()
+
+    promptForValue: (initialValue, @onOk) =>
+      @$el.find('input').val(initialValue)
+      @$el.modal('show')
+      
+    ok: () =>
+      value = @$el.find('input').val()
+      @$el.one('hidden.bs.modal', () => @onOk(value))
```
dak: `@onOk()` appears to be trying to call the 'onOk' method on this View, which doesn't seem to exist?  I would expect this to throw an undefined error if it ever gets executed.

src/scripts/modules/media/tabs/contents/toc/modals/section-name/section-name.coffee
RoyEJohnson: CS automatically translates == to ===. It's exactly the same as is.

src/scripts/modules/media/tabs/contents/toc/modals/section-name/section-name.coffee
RoyEJohnson: @onOk is set in the promptForValue method, which is what shows the dialog.

PR #1180, opened 2015-09-14
Fix error for undefined object

amberwebb:


src/scripts/models/search-results.coffee
```diff
@@ -56,7 +56,8 @@ define (require) ->
 
     fetch: () ->
       fetchArgs = arguments[0]
-      fetchArgs.timeout = 60000 # one minute
+      if fetchArgs isnt undefined
+        fetchArgs.timeout = 60000 # one minute
```
dak: This could be written as `fetchArgs?.timeout = 60000`, or perhaps even better, set a default value to assign to the args:
`fetch: (fetchArgs = {timeout: 60000}) ->`

PR #1182, opened 2015-09-14
Fix padding in featured books more links and notes in content 

amberwebb:
When viewed in portrait mode on mobile

### DIFF
--- a/src/scripts/modules/media/body/body.less
+++ b/src/scripts/modules/media/body/body.less
@@ -40,7 +40,7 @@
     @import 'content.less';
   }
 
-   cnx-pi {
+  cnx-pi {
     display: none;
   }
 
--- a/src/scripts/modules/media/body/body.less
+++ b/src/scripts/modules/media/body/body.less
@@ -40,7 +40,12 @@
     @import 'content.less';
   }
 
-  cnx-pi {
+   cnx-pi {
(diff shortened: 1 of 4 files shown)


src/scripts/modules/media/body/body.less
```diff
@@ -40,7 +40,12 @@
     @import 'content.less';
   }
 
-  cnx-pi {
+   cnx-pi {
```
dak: I don't think this space was intentional?

src/scripts/modules/media/footer/footer.less
```diff
@@ -4,6 +4,8 @@
 .media-footer {
   margin: 4rem 6rem 6rem 6rem;
 
+  @media
```
dak: Is this intended to be here?

dak:
The build failure on the push appears to be because it's building on travis' legacy hardware, as opposed to the pr build which is in a container.  Not sure why that may be the case (maybe rebase from master?), but either way it appears that it can be ignored.

src/scripts/modules/media/footer/footer.less
amberwebb: I pulled this code out from the stash from a month or so ago. This is where I left off, so no it was not intended to be there like that.

PR #1183, opened 2015-09-15
Add keyboard controls to ToC

dak:


src/scripts/modules/media/tabs/contents/toc/section.coffee
```diff
@@ -48,6 +49,12 @@ define (require) ->
       else
         @model.set('expanded', true)
 
+    toggleSectionWithKeyboard: (e) ->
+      if e.keyCode is 13 or e.keyCode is 32
```
pumazi: What keys are 13 and 32?

amberwebb: Enter and spacebar

PR #1191, opened 2015-09-21
Accessibility shortcuts

RoyEJohnson:
Links and buttons participate in tab order. Some had no styling change when they got focus, so I applied stylings, trying to keep them minimal.

### DIFF
--- a/src/scripts/modules/media/footer/footer-template.html
+++ b/src/scripts/modules/media/footer/footer-template.html
@@ -1,13 +1,13 @@
 <div class="media-footer">
   <ul>
     <li class="tab" data-content="downloads">
-        <a class="tab-title" href="#">Downloads</a>
+      <span class="tab-title" tabindex="0">Downloads</span>
     </li>
     <li class="tab" data-content="history">
-      <a class="tab-title" href="#">History</a>
+      <span class="tab-title" tabindex="0">History</span>
     </li>
     <li class="tab" data-content="attribution">
-      <a class="tab-title" href="#">Attribution</a>
+      <span class="tab-title" tabindex="0">Attribution</span>
     </li>
   </ul>
 
--- a/src/scripts/modules/media/footer/footer-template.html
+++ b/src/scripts/modules/media/footer/footer-template.html
@@ -1,13 +1,13 @@
 <div class="media-footer">
   <ul>
     <li class="tab" data-content="downloads">
-      <span class="tab-title">Downloads</span>
+        <a class="tab-title" href="#">Downloads</a>
     </li>
     <li class="tab" data-content="history">
-      <span class="tab-title">History</span>
+      <a class="tab-title" href="#">History</a>
     </li>
     <li class="tab" data-content="attribution">
-      <span class="tab-title">Attribution</span>
+      <a class="tab-title" href="#">Attribution</a>
     </li>
   </ul>
 
--- a/src/scripts/modules/media/footer/footer-template.html
+++ b/src/scripts/modules/media/footer/footer-template.html
@@ -1,13 +1,13 @@
 <div class="media-footer">
   <ul>
-    <li class="tab" data-content="downloads">
-      <span class="tab-title" tabindex="0">Downloads</span>
+    <li class="tab" data-content="downloads" tabindex="0">
+      <span class="tab-title">Downloads</span>
     </li>
-    <li class="tab" data-content="history">
-      <span class="tab-title" tabindex="0">History</span>
+    <li class="tab" data-content="history" tabindex="0">
+      <span class="tab-title">History</span>
     </li>
-    <li class="tab" data-content="attribution">
-      <span class="tab-title" tabindex="0">Attribution</span>
+    <li class="tab" data-content="attribution" tabindex="0">
+      <span class="tab-title">Attribution</span>
     </li>
   </ul>
 
--- a/src/scripts/modules/media/footer/footer.coffee
+++ b/src/scripts/modules/media/footer/footer.coffee
@@ -18,6 +18,7 @@ define (require) ->
 
     events:
       'click .tab': 'selectTab'
+      'keydown .tab': 'keySelectTab'
 
     editable:
       '[data-content="downloads"]':
@@ -36,10 +37,14 @@ define (require) ->
       @regions.attribution.show(new AttributionView({model: @model}))
       @regions.license.show(new LicenseView({model: @model}))
 
+    keySelectTab: (e) ->
+      if e.keyCode is 13 or e.keyCode is 32
+        e.preventDefault()
+        $(e.currentTarget).click()
+
     selectTab: (e) ->
       $tab = $(e.currentTarget)
       @switchTab($tab)
-      false
 
     switchTab: ($tab) ->
       if $tab.hasClass('disabled') then return
--- a/src/scripts/modules/media/footer/footer.coffee
+++ b/src/scripts/modules/media/footer/footer.coffee
@@ -39,6 +39,7 @@ define (require) ->
     selectTab: (e) ->
       $tab = $(e.currentTarget)
       @switchTab($tab)
+      false
 
     switchTab: ($tab) ->
       if $tab.hasClass('disabled') then return
--- a/src/scripts/modules/media/footer/footer.less
+++ b/src/scripts/modules/media/footer/footer.less
@@ -37,6 +37,13 @@
           content: "+";
         }
       }
+      
+      &:focus {
+        border-top: 0.1rem solid @brand-primary;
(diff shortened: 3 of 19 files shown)


src/scripts/modules/media/footer/footer-template.html
```diff
@@ -1,13 +1,13 @@
 <div class="media-footer">
   <ul>
     <li class="tab" data-content="downloads">
-      <span class="tab-title">Downloads</span>
+        <a class="tab-title" href="#">Downloads</a>
```
dak: Indentation appears to be off here.

src/scripts/modules/media/footer/footer-template.html
```diff
@@ -1,13 +1,13 @@
...
     </li>
     <li class="tab" data-content="history">
-      <span class="tab-title">History</span>
+      <a class="tab-title" href="#">History</a>
     </li>
     <li class="tab" data-content="attribution">
-      <span class="tab-title">Attribution</span>
+      <a class="tab-title" href="#">Attribution</a>
```
dak: Why make these anchor tags instead of spans with a tabindex?

src/scripts/modules/media/footer/footer.coffee
```diff
@@ -39,6 +39,7 @@ define (require) ->
     selectTab: (e) ->
       $tab = $(e.currentTarget)
       @switchTab($tab)
+      false
```
dak: If this is being used to prevent the browser's default behavior, please use an explicit `e.preventDefault()`.  While returning false does have that affect in jQuery, it means a bit more, and can be less clear when reading the code.

src/scripts/modules/media/footer/footer.less
```diff
@@ -38,7 +38,7 @@
         }
       }
 
-      &.active {
+      &.active, &.active>a {
```
dak: To help with code clarity, we try to put multiple rules on their own lines, ie:
```
&.active,
&.active > a {
```

src/scripts/modules/media/footer/inherits/tab/tab.coffee
```diff
@@ -15,15 +15,22 @@ define (require) ->
...
 
+    stayHere: (e) ->
+      false
+
     toggleMedia: (e) ->
       @media = $(e.currentTarget).data('media')
       @render()
+      # Find the re-rendered element and apply focus
+      @$el.find('.book-page-toggle .btn.active').focus()
+      false
```
dak: Same issue here with returning false.

src/scripts/modules/media/footer/inherits/tab/toggle-partial.html
```diff
@@ -1,8 +1,8 @@
 {{#is type 'application/vnd.org.cnx.collection'}}
   {{#if contents.length}}
     <div class="book-page-toggle btn-group">
-      <span class="btn {{#is mediaType 'application/vnd.org.cnx.collection'}}active{{/is}}" data-media="book">Book</span>
-      <span class="btn {{#is mediaType 'application/vnd.org.cnx.module'}}active{{/is}}" data-media="page">Page</span>
+      <a class="btn {{#is mediaType 'application/vnd.org.cnx.collection'}}active{{/is}}" data-media="book" href="#">Book</a>
+      <a class="btn {{#is mediaType 'application/vnd.org.cnx.module'}}active{{/is}}" data-media="page" href="#">Page</a>
```
dak: You may want to just use `tabindex="0"` for these and then you don't have to prevent the browser's default behavior at all.  It's also a bit more semantically accurate.

dak:
Can you include screenshots of some of these changes in this PR?
It may also be helpful to break this into multiple PRs to allow for easier commenting on each change.

src/scripts/modules/media/footer/footer-template.html
RoyEJohnson: Arbitrary decision. Is there a reason to consider spans preferable?

RoyEJohnson:
Social media icons at the top, when they get focus, get a light blue border
screen shot 2015-09-17 at 9 58 31 am

Footer tabs and all links (including social media links in footer) get the browser-default highlighting (lightblue border on Chrome, dotted line in Firefox)
screen shot 2015-09-17 at 10 00 17 am

Page/Book toggle buttons get lightblue border
screen shot 2015-09-21 at 11 10 27 am

RoyEJohnson:
Footer tab highlight encloses tab, not just text
screen shot 2015-09-21 at 11 17 05 am

amberwebb:
@RoyEJohnson Maybe you already addressed this, but should the footer tabs highlight in the same style as the ToC tabs? I'm just glancing at your PR, so maybe I missed something.

RoyEJohnson:
@amberwebb That is an open question. My bias, as a non-designer, was not to change things that had highlights, and to use default highlighting style as much as possible when adding highlights. I can certainly copy the focus CSS from ToC to footer if you think that's the way to go.

amberwebb:
I'm not sure why the ToC tab highlighting was done the way it was, but whatever way you go, it should be consistent.

src/scripts/modules/media/header/header.coffee
```diff
@@ -104,6 +105,9 @@ define (require) ->
           router.navigate(href, {trigger: false, analytics: true})
       @model.editOrDeriveContent(options, data)
 
+    getBook: (event) ->
+      console.debug("Getting book", event)
```
dak: Is this debug code intended to be in master?

src/scripts/modules/search/results/breadcrumbs/breadcrumbs.less
```diff
@@ -36,7 +36,7 @@
       background-color: @gray-dark;
       opacity: 0;
 
-      &:hover {
+      &:hover, &:focus {
```
dak: Please put multiple rules on different lines to make the code easier to read:
```
&:hover,
&:focus {
```

src/scripts/modules/media/header/header-template.html
```diff
@@ -36,9 +36,9 @@
       <div class="downloads">
         {{#or hasDownloads buyLink}}
           {{#is mediaType 'application/vnd.org.cnx.collection'}}
-            <a class="btn">Get This Book!</a>
+            <button class="btn" tabindex="0" type="submit">Get This Book!</button>
           {{else}}
-            <a class="btn">Get This Page!</a>
+            <button class="btn" tabindex="0" type="submit">Get This Page!</button>
```
dak: I believe `type="submit"` is the default attribute for buttons.

src/scripts/modules/media/footer/footer.less
```diff
@@ -37,8 +37,16 @@
...
+      &:focus {
+        border-top: 0.1rem solid @brand-primary;
+        border-left: 0.1rem solid @brand-primary;
+        border-right: 0.1rem solid @brand-primary;
+        outline: 0;
+      }
 
-      &.active {
+      &.active,
+      &.active>button {
```
dak: Just a style comment, most of the rest of the CSS/LESS uses whitespace between selector symbols, ie: `&.active > button`.

dak:
For the social media icons, what if they changed green on focus/hover?

dak:
Can the footer tab's focus style match the media tab's focus style?

dak:
I know lightblue is the default Safari/Chrome browser focus indicator, but it doesn't really match the rest of our theme.

RoyEJohnson:
I updated the footer tab focus highlighting to match the ToC highlighting the other day. Footer tabs turn dark blue on active, so the border color blends in, but it can still be noticed when focus arrives and leaves.

I made the header social media icons highlight green (same as green buttons), but the footer icons defied all attempts to modify their focus highlighting.

dak:
@RoyEJohnson Any chance you can update the screenshots?

RoyEJohnson:
screen shot 2015-09-23 at 10 21 22 am
screen shot 2015-09-23 at 10 22 04 am
screen shot 2015-09-23 at 10 22 39 am

dak:
screen shot 2015-09-23 at 11 25 27 am
I was thinking something like this. The contrast is a lot easier to see than just changing the border.

RoyEJohnson:
@dak Looks good to me. Done.

src/scripts/modules/media/footer/inherits/tab/tab.coffee
```diff
@@ -15,15 +15,20 @@ define (require) ->
 
     events:
       'click > .book-page-toggle > .btn:not(.active)': 'toggleMedia'
+      'click > .book-page-toggle > .btn.active': 'stayHere'
```
dak: Why attach an event if it's going to be a noop?

PR #1199, opened 2015-10-05
Epub links

RoyEJohnson:
Exclude zero-size files and files in EPUB format.


edwoodward:
This should exclude all files that are zero size.

PR #1203, opened 2015-10-06
Only exclude downloads whose state is not good

RoyEJohnson:
Turns out we don’t want to restrict EPUBs specifically.


dak:
@RoyEJohnson I think this PR should also undo the change that was removing all epubs on parse.

PR #1208, opened 2015-10-07
Handle page uuids

RoyEJohnson:
Nav links put page uuids in URLs
Router handles URLs with page uuids (as well as page numbers)


src/scripts/router.coffee
```diff
@@ -36,7 +36,8 @@ define (require) ->
       @route linksHelper.componentRegEx, 'media', (uuid, version, page, title, qs) ->
         uuid = uuid.toLowerCase()
         uuid = settings.shortcodes[uuid] if settings.shortcodes[uuid]
-        @appView.render('contents', {uuid: uuid, version: version, page: Number(page), qs: qs})
+        pageId = if page?.length > 7 then page else Number(page)
```
dak: Can you add a comment as to why you're checking if the `page.length > 7`?  It's not obvious what the point of that is.

PR #1219, opened 2015-11-10
Handle short id addresses

RoyEJohnson:
URLs display using short id’s instead of uuids for both book and page.

### DIFF
--- a/src/scripts/router.coffee
+++ b/src/scripts/router.coffee
@@ -33,11 +33,12 @@ define (require) ->
         @appView.render('role-acceptance')
 
       # Match and extract uuid and page numbers separated by a colon
-      @route linksHelper.componentRegEx, 'media', (uuid, version, page, title, qs) ->
-        uuid = uuid.toLowerCase()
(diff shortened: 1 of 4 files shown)


src/scripts/router.coffee
```diff
@@ -33,11 +33,12 @@ define (require) ->
...
-      @route linksHelper.componentRegEx, 'media', (uuid, version, page, title, qs) ->
-        uuid = uuid.toLowerCase()
+      @route linksHelper.contentsLinkRegEx, 'media', (uuid, version, page, title, qs) ->
+        # Downcase uuids, not short ids
+        if uuid.length > 10
+          uuid = uuid.toLowerCase()
         uuid = settings.shortcodes[uuid] if settings.shortcodes[uuid]
-        # page numbers will be less than 8 digits; others are uuids
-        pageId = if page?.length > 7 then page else Number(page)
+        pageId = if isNaN(page) then page else Number(page)
```
reedstrm: rather than delete the comment, might be worth mentioning supporting page number urls via if it parses as an int.

PR #1221, opened 2015-11-12
Fixes

RoyEJohnson:
Nav was updating model before changing pages
Separated short Id handler out so it isn’t called every render
node.coffee now uses regexp in linksHelper file; split pattern in
linksHelper so it can do that
finding the page by shortID works with or without version number


reedstrm:
Looks good. Merging.

PR #1222, opened 2015-11-12
Wide viewport no top tabs

RoyEJohnson:
The viewport can be 1260px wide, but media still only 960px
Metadata tab moved to bottom and re-labeled More Information
Tabs from top section removed; Contents now controlled by a toggle
button next to the nav bar control


reedstrm:
Hmm, is this part of the 'can release before TOC changes' code? We're splitting things up, right?

RoyEJohnson:
@reedstrm Right, this should be suitable for individual release. I'm working on the next phase, which will have the TOC on the left side, but will (still) not have the in-book search or new menu.

PR #1224, opened 2015-11-12
Ensures URL for content is canonical

RoyEJohnson:
linksHelper.getPath now returns short-id@version for both book and
page, which covers all navigation links.
Content body checks that the URL matches the canonical path or updates
it if not.


reedstrm:
passed locally. Merging

PR #1232, opened 2015-11-16
Scroll content toc left

RoyEJohnson:
This is a big one, due to lots of restructuring.

1. Table of contents and page content display side by side, scroll
independently
2. Scrolling content down minimizes title area, creating more reading space
onscreen
3. Pages redesigned to use main-page view, which loads footer at the bottom, so they don’t have to load it themselves;
some updates required to navigation controls due to restructuring.
4. Minimal layout is mostly a modified page header now (for content),
so minimal/media hierarchy is gone.
5. Canonical paths updated to remember query string


dak:
@RoyEJohnson There's a lot of transformation code that modifies how pages load in the body as well as styling for elements that appears to be removed in this PR.  Is this correct, and if so, is that code no longer necessary?

RoyEJohnson:
@dak Give me a "for instance" or two so that I'm sure I know what you're referring to. There *are* a lot of changes to how pages load, and a fair amount of code was removed. There certainly could be more now-obviated code that I didn't notice.

dak:
@RoyEJohnson One particular example is the `@renderDom` function that existed in body.coffee that was used to perform numerous transforms necessary to make the HTML sent from cnx-archive/authoring render properly on the web.

src/scripts/modules/media/nav/nav.less
```diff
@@ -9,86 +9,115 @@
...
-  > .book-nav {
-    .flexbox();
-    .flex-flow(row; nowrap);
-    .justify-content(center);
-    .user-select(none);
-
-    > .progress, > .back-to-top {
-      .flex(1; 1; 75%);
+  > .media-toolbar {
+    >* {
```
dak: I'm always a little skeptical whenever using the `*` in CSS.  Is there no better way to do this?

src/scripts/modules/media/nav/nav.less
```diff
@@ -9,86 +9,115 @@
...
+        background-color: white;
+        color: @brand-primary;
       }
     }
 
-    > .back-to-top {
-      text-align: center;
-      line-height: @nav-height;
+    .media-navbar {
+      width: ~"calc(90% - 9em)";
```
dak: Would it be better to use `rem` here instead of `em`?

PR #1234, opened 2015-11-18
Add support to hide exercises for concept coach

dak:


src/scripts/modules/media/body/body.coffee
```diff
@@ -209,6 +209,15 @@ define (require) ->
           # # uncomment to embed fake exercises and see embeddable exercises in action
           # @fakeExercises($temp)
 
+          # Hide Exercsies for Concept Coach
```
RoyEJohnson: Exercises typo

src/scripts/modules/media/body/body.coffee
```diff
@@ -209,6 +209,15 @@ define (require) ->
...
+          if settings.hideExercises.indexOf(@model.getUuid())
+            hiddenClasses = []
+            processingInstructions = @$el.find('.media-body').find('cnx-pi[data-hide]')
```
RoyEJohnson: This could be a single find, couldn't it?

dak: Multiple finds is actually more performant.  CSS selector searching works backwards, so `find('.media-body cnx-pi[data-hide]')` would first find every single element that matches `[data-hide]`, then of those match `cnx-pi`, then of those match `.media-body`.  Restricting the search set a bit first using a class lookup helps a bit.

PR #1237, opened 2015-11-18
In book search

RoyEJohnson:
Adds "Search this book" search bar to navigation bar
Search results are displayed in TOC, which is filtered to show only the matched pages and their containing sections.
Content body shows page with matches highlighted


dak:
@RoyEJohnson This needs to be rebased so it can be merged.

RoyEJohnson:
Ok, I think this is finally ready. Note that it depends on a new proxy pass entry for /contents/search.

RoyEJohnson:
Ok, NOW it's ready.

src/scripts/modules/media/tabs/contents/toc/section.coffee
```diff
@@ -24,37 +26,31 @@ define (require) ->
...
-
       _.each nodes, (node) =>
-        if node.isSection()
-          @regions.container.appendAs 'li', new TocSectionView
-            model: node
-        else
-          @regions.container.appendAs 'li', new TocPageView
-            model: node
-            collection: @model
+        if node.get('visible') ? true
```
dak: This is quite clever, but not really the easiest code to follow.  If you want to default to the `visible` property being `true`, why not just add that to the node model?

PR #1262, opened 2015-12-01
Change header hiding

RoyEJohnson:
Only show header when scrolled to (almost) top. Hide header when scrolled down a portion of the window height. When hiding header, reduce title size.

### DIFF
--- a/src/scripts/modules/media/media.coffee
+++ b/src/scripts/modules/media/media.coffee
@@ -28,9 +28,8 @@ define (require) ->
   ###
   headerController = ($el, selector, compactSelector) ->
     threshold = window.innerHeight / 5
-    fudge = 8
     hideAbove = threshold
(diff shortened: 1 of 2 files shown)


src/scripts/modules/media/media.coffee
```diff
@@ -23,26 +23,27 @@ define (require) ->
...
   a scroll event to hook it up.
   ###
-  headerController = ($el, selector) ->
-    oldTop = 0
-    inTransition = false
+  headerController = ($el, selector, compactSelector) ->
+    threshold = window.innerHeight / 5
+    fudge = 8
+    hideAbove = threshold
+    showBelow = fudge
```
dak: What's the point of the `fudge` variable if it's just being set to `showBelow` and then never used again?

PR #1268, opened 2015-12-03
fix css non-loading through less loader

pandafulmanda:
different behavior for dist vs local, fixed and should work both ways...hopefully.  Works locally when `npm run dist`


reedstrm:
@dak, @roy, could you verify that this is the way to go for loading both in dev and dist?

PR #1275, opened 2015-12-04
Scrolling redesign

RoyEJohnson:
Moved scrolling from content divs back to window.
Content header is “pinned” to top of window when scrolled that far
TOC stretches from content header to bottom of viewport, scrolls
independently.


RoyEJohnson:
I haven't tested this enough yet.

PR #1287, opened 2015-12-07
Fix pinned header problems

RoyEJohnson:
Skip to header was messing things up. Absolute-positioning its div fixed that.
See commits for other bug fix descriptions.

### DIFF
--- a/src/scripts/modules/media/body/body-template.html
+++ b/src/scripts/modules/media/body/body-template.html
@@ -16,5 +16,6 @@
         <span class="sr-only">Loading</span>
       </div>
     </div>
+    <div class="spacer"></div>
   </div>
 {{/if}}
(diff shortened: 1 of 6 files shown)


src/scripts/modules/media/title/title.coffee
```diff
@@ -39,6 +39,9 @@ define (require) ->
       @listenTo(@model, 'change:loaded change:title change:canPublish', @render)
       @listenTo(router, 'navigate', @render)
 
+    onRender: ->
+      @trigger('render')
```
dak: What is listening to this event?

src/scripts/modules/media/body/body-template.html
```diff
@@ -16,5 +16,6 @@
         <span class="sr-only">Loading</span>
       </div>
     </div>
+    <div class="spacer"></div>
```
dak: This looks hacky.

reedstrm: my understand of it is to hint to the browser to save space for the header that is about to be rendered, this avoiding a flash/reflow. So, yes, a hack, sort of. Got a better way?

PR #1296, opened 2015-12-09
Attempt to hack a toggle for Zendesk

dak:
This is an extremely ugly fishing expedition for a hack to allow Zendesk to be toggled.  I have no idea if it actually works, and I'm not sure there's a good way to test it other than inputing feedback and seeing where it goes.


reedstrm:
I'm assuming zendesk.js is taken straight from Zendesk? Is there some way to tag it so as not to run jshint against it?

PR #1311, opened 2015-12-14
Stop polling profile

RoyEJohnson:
Attempt to load profile once per page


reedstrm:
Could we add polling when the page gets focus, like changing tabs?

dak:
This should be easily implemented.  See: https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API

PR #1312, opened 2015-12-14
Poll for user profile on window focus

RoyEJohnson:
If user goes to another tab or window or application and comes back.

### DIFF
--- a/src/scripts/modules/header/header.coffee
+++ b/src/scripts/modules/header/header.coffee
@@ -8,6 +8,8 @@ define (require) ->
   require('bootstrapCollapse')
   #require('zendesk')
 
+  $(window).focus(session.update.bind(session))
+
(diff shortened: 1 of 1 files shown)


dak:
I think this should be using the new [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) instead.  My understanding is that window.focus is non-standard and we shouldn't assume support will continue (although it does exist in all browsers at the moment).

reedstrm:
http://caniuse.com/#feat=pagevisibility Hmm, seems all the most recent support it.

RoyEJohnson:
@dak I don't find anything that says window focus events are nonstandard or going away. Focus changes in more situations than visibility (as the visibility API page notes). In particular, you could go to another window without minimizing the first, and visibility would not be triggered.

PR #1316, opened 2015-12-15
Fix uuid redirection

RoyEJohnson:
Rewrote page lookup so it finds page uuids without version. Had it set error to 404 if page is not found.


RoyEJohnson:
Ok, I'm really done tweaking this. Pinky swear.

PR #1330, opened 2015-12-21
clean mathjax rendering code

pandafulmanda:


src/scripts/modules/media/body/body.less
```diff
@@ -64,8 +64,12 @@ body.cc-opened {
   .concept-coach-wrapper {
     text-align: left;
   }
+
   [data-math]:not(.math-rendered) {
-    display: none;
+    visibility: hidden;
+    .MathJax {
```
philschatz: Is this so the MathJax preview shows up?

pandafulmanda: yes

philschatz: "why hide the parent but not the child?"

And could you include a comment in the CSS? I don't think it would be clear looking back at the code in a couple months

pandafulmanda: okay, i think im just going to take out the `.MathJax{}` part

PR #1334, opened 2016-01-01
Coach with urls

pandafulmanda:
includes:
* opening coach changes urls
  * such that when you click back while coach is opened, it closes the coach and leaves you back on the same page you launched the coach from
  * cnx nav links ignore the coach view query parameter by default, so that the cnx page nav for back and next navigates to the destination page with the coach closed
* taking the coach mounting out of a deferred
* clean up the `isCoach` -- no need for it to be a template helper anymore

Will be followed up by an update to the coach version after openstax/concept-coach#104 is merged, although this PR can be merged independent of a coach update

### DIFF
--- a/src/scripts/modules/media/body/body.coffee
+++ b/src/scripts/modules/media/body/body.coffee
@@ -64,7 +64,7 @@ define (require) ->
       else
         pageId = 0
       currentRoute = Backbone.history.getFragment()
-      canonicalPath = linksHelper.getPath('contents', {model: @model, page: pageId})
+      canonicalPath = linksHelper.getPath('contents', {model: @model, page: pageId}, [])
       if (canonicalPath isnt "/#{currentRoute}")
         router.navigate(canonicalPath, {replace: true})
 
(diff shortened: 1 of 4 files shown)


src/scripts/modules/media/body/body.coffee
```diff
@@ -85,8 +76,24 @@ define (require) ->
       else
         $els.hide()
 
+    goToPage: (pageNumber, href) ->
+      @model.setPage(pageNumber)
+      router.navigate href, {trigger: false}, => @parent.parent.parent.trackAnalytics()
+
+    getCoach: ->
+      moduleUUID = @model.getUuid()?.split('?')[0]
+      settings.conceptCoach?.uuids?[moduleUUID]
```
RoyEJohnson: Is there a chance moduleUUID will be null and it will try to index with it?

kathi-fletcher:
I wasn't sure if the following applies to the browser back or just to the cnx nav "back" -- "links ignore the coach view query parameter by default, so that the page nav for back and next navigates to the destination page with the coach closed"

If I go to page A, then A-coach, then B, and then use browser back button, do I go to A or A-coach? I think that maybe the above implies that going to B will also close A-coach, and therefore back goes to A (which would be the desired behavior). But I am not quite sure.

src/scripts/modules/media/body/body.coffee
pandafulmanda: `settings.conceptCoach.uuids` is an object, so no.  Also, looking into the [`getUuid` function](https://github.com/Connexions/webview/blob/master/src/scripts/models/contents/node.coffee#L191), it will only return strings, so `moduleUUID` would be an empty string in the most "falsiest" case.

pandafulmanda:
### Common Path
Going from page, opening coach, closing coach, clicking next page
`A` → `A-coach` → `A` → `B`
**Clicking on browser back**: `A`

PR #1336, opened 2016-01-04
Handle no-pages condition

### LINKED ISSUE #1335: Books without pages display 404
As reported by Dick Baldwin, versions of books without pages, such as:

http://cnx.org/contents/c9c9bdca-4281-4152-b347-d891e886a3dc@1.1

display a 404. Webview could handle this case, by using a special error code when there's no page in the collection json, then special-case rendering that (give access to book metadata, but a 'This book contains no pages' message)

RoyEJohnson:
If book contains no pages, don’t 404 in setPage, just return.

Fixes #1335

### DIFF
--- a/src/scripts/models/content.coffee
+++ b/src/scripts/models/content.coffee
@@ -106,6 +106,7 @@ define (require) ->
         @trigger("change:currentPage.#{eventName.slice(7)}", page, value)
 
     _setPage: (page) ->
+      return if @cachedPages.length == 0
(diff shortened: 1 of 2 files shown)


reedstrm:
could you remove the changes to settings.js? And what does this actually then display?

reedstrm:
Let's merge this to dev as a minimal fix, then we'll see about adding something to the page area.

PR #1345, opened 2016-01-07
Encode query terms

### LINKED ISSUE #1343: Title search fails when it includes a question mark
Original ticket: Connexions/cnx-archive#301

----

@edwoodward:
I did a search for the title "What Is Consciousness?" The page blinked, but did nothing. I removed the question mark and it worked as expected.

@karenc:
This appears to be a webview bug. I went to the advanced search page and did a title search for "what is consciousness?", it went to http://cnx.org/search?q=title:%22what%20is%20consciousness?%22 but instead of displaying the search results, it showed the advanced search page again.

If I escape the question mark by hand to http://cnx.org/search?q=title:%22what%20is%20consciousness%3F%22, the search results page shows correctly.

RoyEJohnson:
URI encode search terms so that question marks and such work
Fixes #1343

### DIFF
--- a/src/scripts/modules/minimal/search/advanced/advanced.coffee
+++ b/src/scripts/modules/minimal/search/advanced/advanced.coffee
@@ -40,6 +40,7 @@ define (require) ->
       format = (obj) ->
         _.map obj, (limit, index) ->
           if not limit.value then return
+          limit.value = encodeURIComponent(limit.value)
 
           # Split keywords into multiple keyword:value pairs
           if limit.name is 'keywords'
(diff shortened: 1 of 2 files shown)


reedstrm:
So this is a display-results-time fix, not changing what's sent to archive, correct?

RoyEJohnson:
No, this formats the query string for use in a URL. The process is: user fills out form, submits, query gets turned into URL, navigation parses the URL and makes the search request. This is part of turning it into an acceptable URL (problem was noticed when a ? was in the query).

reedstrm:
further poking at production shows that archive is sensitive to decoding %XX URLs and probably needs some input santization: e.g. http://archive.cnx.org/search?q=title:%centage  throws an error.

PR #1346, opened 2016-01-07
Handle scroll-to-anchor in content

### LINKED ISSUE #1344: Archor links are under header
This was reported by Richard Baldwin

Go to http://cnx.org/contents/mNJTwIBE@35.51:c-n7XnXf@5/Dsp00095-Preface-to-Digital-Si

Open the Contents

Select the item that reads "3 Plotting Data"

Select 
3.1 Java1468-Plotting Engineering and Scientific Data using Java         

Under "Table of contents", select Listings

Note that the view does not land on the local anchor that reads "Listings". Instead it lands about four lines down the page from that anchor on the line that reads:

Listing 3 . The method named getNmbr.

RoyEJohnson:
Pinned navbar obscures top of window, so when going to id, scroll back
by the height of the bar.
Removed some old scrolling code for element that no longer scrolls
Fixes #1344

### DIFF
--- a/src/scripts/modules/media/media.coffee
+++ b/src/scripts/modules/media/media.coffee
@@ -98,10 +98,18 @@ define (require) ->
         newHeight = window.innerHeight - tocTop
         $toc.height("#{newHeight}px")
       Backbone.on('window:optimizedResize', setTocHeight)
+      adjustHashTop = ->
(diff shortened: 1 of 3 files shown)


reedstrm:
make sense

PR #1360, opened 2016-01-27
Rollback cc

reedstrm:
This should rollback the CC hash to what is currently deployed on production, to allow for rollout of other webview fixes, including google analytics codes, and scroll to target.


reedstrm:
Don't want this on master, actually. Closing in favor of #1361

PR #1375, opened 2016-02-04
tracker name for GA must be alphanumeric

pandafulmanda:
# GA tracker doesn't like hyphens in the trackerName

Used [ga-debugger](https://developers.google.com/analytics/devguides/collection/analyticsjs/debugging) to trace problem

## Before
screen shot 2016-02-04 at 3 12 52 pm

## After
screen shot 2016-02-04 at 3 13 33 pm

# also, get rid of lessc's warning about [XHR sync](https://github.com/Connexions/webview/pull/1375/files#diff-722f26ce6d75fb7d146f54a4ce91810cR208)

Only affects dev version, wasn't causing problems currently, but this error broke less rendering in sims repo for example in the latest Chrome

`require-less` tries to use XHR `async=false` option by default, this [no longer follows spec](https://xhr.spec.whatwg.org/#sync-warning)

screen shot 2016-02-04 at 3 15 15 pm

### DIFF
--- a/src/scripts/helpers/handlers/analytics.coffee
+++ b/src/scripts/helpers/handlers/analytics.coffee
@@ -30,7 +30,9 @@ define (require) ->
       require ['analytics'], (ga) ->
         # Use the default analytics ID in settings if no account is specified
         account ?= settings.analyticsID
-        ga('create', account, 'auto', account)
+        # TODO investigate if we need a different name for our tracker name
+        trackerName = 'cnxTracker'
+        ga('create', account, 'auto', trackerName)
 
-        ga("#{account}.send", 'pageview', fragment)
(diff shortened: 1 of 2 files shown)


src/scripts/helpers/handlers/analytics.coffee
```diff
@@ -30,7 +30,9 @@ define (require) ->
       require ['analytics'], (ga) ->
         # Use the default analytics ID in settings if no account is specified
         account ?= settings.analyticsID
-        ga('create', account, 'auto', account)
+        # TODO investigate if we need a different name for our tracker name
```
RoyEJohnson: Is this the sort of thing we should create Issue for?

PR #1379, opened 2016-02-05
Update CC build with latest fixes

nathanstitt:
* Remember free response when switching between exercises
 * Screen to update student identifier


nathanstitt:
Don't merge just yet, i'm updating to use a Tag

PR #1388, opened 2016-02-23
adds source maps

### LINKED ISSUE #1378: Add sourcemaps to all builds (including production)
This would allow the dev instance to run much faster and allow debugging on production.

The code is already open-source and the _only_ overhead is a 10-character comment at the end of the minified JS file.

Plus, it could help get people on the internet to fix bugs or maybe learn some coding :smile:

related: openstax/tutor-js#916

philschatz:
image

The files load but I am not sure about the best places to set breakpoints to verify. Also, I did not look very hard to see if `requirejs` could include the original `.coffee` source (see screenshot)

Fixes #1378

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -138,6 +138,7 @@ module.exports = (grunt) ->
           keepBuildDir: false
           preserveLicenseComments: false
           skipDirOptimize: true
+          generateSourceMaps: true
           optimize: 'none'
           stubModules: ['cs']
           modules: [{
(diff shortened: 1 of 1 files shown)


reedstrm:
What does this screenshot demonstrate?

philschatz:
It shows that rather than showing the mangled (minified) Javascript, the source mapping document is loaded. The grey background on some lines means you can no longer break on that line (probably because the mangling reordered some statements or something)

reedstrm:
Ah, so we'd switch dev to run in a minified mode - right  forgot that step.

philschatz:
Yep, that makes dev run more like production, and the source maps should also be on production. I think someone (maybe @dak or @RoyEJohnson ) might want to verify the source code actually still lines up (by setting breakpoints and stepping through or something)

RoyEJohnson:
I've set a few breakpoints and they behave as expected.

philschatz:
:sparkles: Yay! :sparkles: should I post somewhere on the deployment-side of things to also run minified for `dev`?

PR #1389, opened 2016-02-23
WIP - initial loading coach remotely

pandafulmanda:


src/scripts/config.js
```diff
@@ -1,6 +1,9 @@
 (function () {
   'use strict';
 
+  var coachBasePath = '//openstax.github.io/concept-coach/';
```
philschatz: - just verifying that this URL is for testing purposes only, right?
- will this URL be loaded for everyone that goes to CNX or just those in a CC class?

pandafulmanda: * yes
* i will need to check, and we can look into doing that, but i think that's secondary to getting coach to not need a hash update in cnx right?

philschatz: I believe CNX has a policy of not depending on other sites (except google analytics and I guess now zendesk) but others might be able to chime in.

edwoodward:
@pandafulmanda Is this still being worked on?

pandafulmanda:
we will eventually get back to this, but we can close this for now

PR #1391, opened 2016-02-26
prevent Cannot read property 'top' of undefined errors

pandafulmanda:
from when hash changed query is not found.  This would happen whenever hash events fire and it's `#` or even when it's a valid hash change but the element is not found.

screen shot 2016-02-26 at 3 48 25 pm


Dantemss:
:tada: yay

PR #1401, opened 2016-03-22
Summary toggle shows only when abstract summary exists.

darrenmason:
currentPage.abstract retained a cnxml wrapper which caused the header Summary toggle to display even when no abstract was present.

### DIFF
--- a/src/scripts/models/contents/node.coffee
+++ b/src/scripts/models/contents/node.coffee
@@ -38,7 +38,7 @@ define (require) ->
       #if @get('title') then delete response.title
       response.title = @get('title') or response.title
 
-      # extract abstract from cnxml wrapper
+      # extract abstract from wrapper
       if response.abstract?
         response.abstract = $(response.abstract).unwrap().html()
 
--- a/src/scripts/models/contents/node.coffee
+++ b/src/scripts/models/contents/node.coffee
@@ -38,6 +38,10 @@ define (require) ->
       #if @get('title') then delete response.title
       response.title = @get('title') or response.title
 
+      # extract abstract from cnxml wrapper
+      if response.abstract?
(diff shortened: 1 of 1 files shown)


reedstrm:
The description confuses me slightly - there should be no CNXML at this point, if there is, it's a bug
in the CNXML to HTML transforms. Poking at some examples (you have the bug tag on this, but I see no associated issue, BTW) I do see that there's an HTML div wrapper around all the abstracts, including empty ones. It does carry a whole bunch of extra namespaces (including CNXML) but that's harmless (and is a bug).  This change unconditionally drops the wrapper.  I'v checked an example doc with a fairly long abstract, http://cnx.org/contents/a2kjFUPm@3 and it changes the display, since it loses the block formatting - the test is now unconstrained on the right. I'd suggest two things - file a bug against rhaptos.cnxmlutils about putting an empty div in when the abstract is empty, and changing this test to see if the div has children.

reedstrm:
I take that back. I accidentally edited out two divs, not just the one removed in this PR. There seems to be no actual visible affect.

src/scripts/models/contents/node.coffee
```diff
@@ -38,6 +38,10 @@ define (require) ->
       #if @get('title') then delete response.title
       response.title = @get('title') or response.title
 
+      # extract abstract from cnxml wrapper
```
reedstrm: could you change the comment to not say cnxml, since this is misleading?

edwoodward:
@darrenmason Can you respond to Ross' comment so we can get this merged?

reedstrm:
I could change the comment myself and merge this ...

darrenmason:
@reedstrm I did this on sustain so I was not entirely sure of the terminology, sorry for the confusion.  Basically there was a wrapper that prevented selection.

PR #1406, opened 2016-05-25
Fix/ keep minimal in new version url

Mayu-TM:
Fixed latest.coffee so that when the url to the new version is clicked from minimal view, it will maintain the minimal view in the new window.

### DIFF
--- a/src/scripts/modules/media/footer/footer.coffee
+++ b/src/scripts/modules/media/footer/footer.coffee
@@ -13,7 +13,7 @@ define (require) ->
     template: template
 
     templateHelpers:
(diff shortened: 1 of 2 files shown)


src/scripts/modules/media/footer/footer.coffee
```diff
@@ -13,7 +13,7 @@ define (require) ->
     template: template
 
     templateHelpers:
-      minimal: () ->
+      minimal:() ->
```
RoyEJohnson: I think you want the space here. Just for stylistic consistency.

PR #1407, opened 2016-05-27
Add Jump to Concept Coach button

dak:
Screenshot

### DIFF
--- a/src/scripts/modules/media/body/body.coffee
+++ b/src/scripts/modules/media/body/body.coffee
@@ -416,7 +416,9 @@ define (require) ->
         @parent?.regions.self.append(new SimModal({model: @model}))
 
       # mount the Concept Coach if the mounter has been configured/if `canCoach`
-      @regions.coach.append(new Coach({model: @model})) if @isCoach()
+      if @isCoach()
+        @coach = new Coach({model: @model})
+        @regions.coach.append(@coach)
 
       # MathJax rendering must be done after the HTML has been added to the DOM
       MathJax?.Hub.Queue =>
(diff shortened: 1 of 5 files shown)


philschatz:
Can you include a screenshot in the PR body? That helps UX (& QA) have a way to verify and/or discuss the changes

philschatz:
Why not just make the "Jump to Concept Coach" a link, and add an `id='concept-coach'` to the SVG image on the bottom of the page?

dak:
There's 2 issues with the link only suggestion:

1) From the story: `Behavior of button is the same as the “Launch Concept Coach” button at the bottom of the reading.`. That to me suggests the button needs to actually launch concept coach, not just move the user to it on the screen.

2) It is explicitly stated it needs to be functional on intro pages too, which don't have the concept coach widget on them (so the SVG image doesn't exist on the page to jump to).

dak:
Screenshot

philschatz:
oh, I might have misunderstood the ticket & thought it was meant to scroll down; but I agree, if the intro modules don't have any exercises I'm not sure what does/should happen in that case. /cc @jason-holmes @Fredasaurus @rvnewell (`@openstax/ux`)

jason-holmes:
We had originally intended the big button at the top to launch CC directly, but that behavior was later changed to scroll down the page to the launcher area instead...sorry the documentation wasn't updated.  The correct behavior for that top button is to scroll down the page to the launcher graphic.

Re: what to do with page modules in CC books that don't have questions (eg, Introductions).  I just discussed this with Kim...for now, we'd like to remove both the big top button AND the launcher from these pages.

dak:
@jason-holmes For pages in CC books without the launcher (Introductions), should the "Get This Book" button still be visible, or should no button be visible there at all (no 'Jump To CC' or 'Get This Book button').

jason-holmes:
“Get this book” button should NOT be visible for CC books.

PR #1411, opened 2016-06-06
Fix/filter languages

Mayu-TM:
Filters the languages on the advanced search options so that only the languages in which we have content is displayed. The languages are sorted in alphabetical order.

### DIFF
--- a/src/scripts/collections/languages.coffee
+++ b/src/scripts/collections/languages.coffee
@@ -14,9 +14,10 @@ define (require) ->
       availableLanguages = {}
       codes = []
       _.each languages, (language) ->
+        code = language[0].substring(0, 2)
         lang =
-          code: language[0].substring(0, 2)
-          name: allLanguages["#{language[0].substring(0, 2)}"].english
+          code: code
+          name: allLanguages[code].english
         codes.push(lang)
 
       # Sort the languages in alphabetical order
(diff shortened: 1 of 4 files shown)


src/scripts/collections/languages.coffee
```diff
@@ -0,0 +1,31 @@
...
+    url: "#{archive}/extras"
+
+    parse: (response) ->
+      languages = response.languages_and_count
+      allLanguages = settings.languages
+      availableLanguages = {}
+      codes = []
+      _.each languages, (language) ->
+        lang =
+          code: language[0].substring(0, 2)
```
RoyEJohnson: In general, avoid repeating expressions. Assign to a variable.

src/scripts/collections/languages.coffee
```diff
@@ -0,0 +1,31 @@
...
+          name: allLanguages["#{language[0].substring(0, 2)}"].english
+        codes.push(lang)
+
+      # Sort the languages in alphabetical order
+      codes.sort (a,b) ->
+        return if a.name > b.name then 1 else -1
+
+      for lang in codes
+        availableLanguages["#{lang.code}"] = allLanguages["#{lang.code}"]
```
RoyEJohnson: Stringifying `lang.code` isn't necessary. It's already a string.

PR #1412, opened 2016-06-08
Fixes the issue of numbering for questions not showing up

Mayu-TM:
Changed content.less so that the numbering shows up for the Review and Discussion questions at the end of the module.


Mayu-TM:
This has errors that I didn't notice at first.

PR #1438, opened 2016-07-27
override list styles for coach so that they have single numbers

pandafulmanda:
# Fixes [weird coach list formatting](https://www.pivotaltracker.com/story/show/126243975)

## from this:
screen shot 2016-07-27 at 5 46 00 pm

## to this:
screen shot 2016-07-27 at 5 24 04 pm

# However

I don't think this is necessarily the right way to solve this problem/maybe I don't understand the full scope behind the list styles defined [here](https://github.com/Connexions/webview/blob/master/src/scripts/modules/media/body/content.less#L182-L365).

For example, if the purpose is to have custom control over the enumeration of `ol`s then why is this [`list-style-type`](https://github.com/Connexions/webview/blob/master/src/scripts/modules/media/body/content.less#L309) not set to `none` since the [`::before`](https://github.com/Connexions/webview/blob/master/src/scripts/modules/media/body/content.less#L316-L321) will be enumerating?

Or, instead, should these [selectors](https://github.com/Connexions/webview/blob/master/src/scripts/modules/media/body/content.less#L351-L352) be a little more strict and be something like `ol[data-display]:not([data-display='inline'])` such that only lists with `[data-display]` attribute be customized?

Either of these changes that affect webview more widely would make this PR unnecessary.

### DIFF
--- a/src/scripts/modules/media/body/body.less
+++ b/src/scripts/modules/media/body/body.less
@@ -95,6 +95,7 @@ body.cc-opened {
     visibility: hidden;
   }
 
+  // various overrides of webview styles for concept-coach
   .media-body > #content .concept-coach-wrapper {
     .dropdown-menu > li > a {
       text-decoration: none;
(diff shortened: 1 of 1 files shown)


RoyEJohnson:
Sorry, I merged this before noticing the question status. I've reverted.

pandafulmanda:
who should I talk to about this?

RoyEJohnson:
It looks like Phil Schatz did a lot of the numbering spec. Maybe start with him?
https://github.com/Connexions/webview/commit/d6fb1d5f7e215f8393d00ea54a973b1055792f13

PR #1448, opened 2016-08-12
don't number footnote list w/ CSS

reedstrm:
Make the CSS for suppressing footnote numbers work. (style: none)


reedstrm:
Example bad page: http://cnx.org/contents/Ph_ExrCQ@8/Experimental-Design-and-Ethics

(footnote list near bottom)

PR #1458, opened 2016-09-30
Webview internationalization

ludekarts:
This pull request enables the OpenStax Webview website to be translated and displayed in the user’s preferred language.
### L20n introduction

The majority of the site is internationalized using L20n, a library for internationalizing HTML/Javascript. It detects user’s preferred language from browser settings, using English as fallback. L20n provides support for 170 languages, and support for additional languages can be added by just defining pluralization and gender-related rules for that language. Compatible language codes can be found in _l20n.js_ file.
### Internationalization

In our internationalization approach we divided Webview into logical chunks as seen from the user’s perspective. The chunks contain the content appearing on specific pages, as well as in 14 other sections, e.g., Main Page, Search, About Us. The division is reflected in _/src/locale/[language-code]/dictionary.ftl_ files. This structure makes it easier for translators to navigate through the file and also helps them get a general idea about the context.

The strings are translated by adding L20n attribute `<data-l10n-id="string_key" data-l10n-args="{key : value}">` to HTML tags. The tags are located next to the default Webview content. This solution guarantees that if L20n fails for some reason, Handlebars templates should render their contents normally.

A small part of the site content cannot be internationalized directly using L20n tags in HTML. This content, such as the strings for the donation slider, is located within CoffeeScript scripts. Translations of those strings were also placed in the .ftl locale files. We based internationalization of this content on L20n’s awareness of DOM changes. This made changes to the code straightforward (in most cases), by just supplying the dynamic generated HTML content with L20n attributes as translation hooks. In some critical cases we used L20n’s Javascript API directly, however it is not the desired approach. More details about translating CoffeeScript can be found in [CNX Internationalization Guidelines for Developers](https://docs.google.com/document/d/19EY3YFrcVdbXC0KMFK8QCSqvYRXg_TtvbaNcUiDwXUI/edit).

Currently, asset localization is handled outside the L20n until it gets fully supported by the framework. Currently each aspect of the Webview model is provided with _$locale_ variable containing current locale code. We also extended function of views so they can be automatically re-rendered when the preferred language setting is changed. Example assets are the images on the “About us” page. Internationalization of these images is handled by linking to localized images in respective locale folders. The files authoring.svg, learning.svg, and map.png files are located in _/src/locale/[language-code]/images/_.

The internationalization solution also takes care of displaying dates in the format consistent with the browser's selected language.
### L20n patch

During development we stumbled upon a problem with the Safari browser which provides locale ID for the current locale ([lang. designator] - [lang. Region], e.g., en-US) rather than just language designators (ISO 639-1). This causes L20n to fallback to the declared default language when locale xx-yy is not found. L20n treats all language codes (xx, xx-yy, xx-yy-zz) as separate instances. There is no default fallback mechanism for codes within the same locale namespace. Consequently, supporting browsers that use locale ID would require duplicating the same translated content in two locale files xx and xx-yy.

To address those issues we authored a patch for L20n. It provides basic language negotiation algorithm to detect ‘closest’ language for the user. As the patch is currently waiting to be reviewed by the L20n dev team, we put the patched L20n.js in _/dist/scripts/patch/_, so that it could be removed and replaced by L20n bower package once the patch is formally accepted.


src/scripts/helpers/backbone/views/base.coffee
```diff
@@ -71,8 +74,14 @@ define (require) ->
       currentPage = @model.getPageNumber().toString() if @model? and @model.isBook?()
       if Backbone.history.fragment?.match? and linksHelper.getCurrentPathComponents().page?
         historyPage = linksHelper.getCurrentPathComponents().page
+      title = document.querySelector('title')
       if @pageTitle
-        document.title = linksHelper.stripTags(@pageTitle) + settings.titleSuffix
+        title.dataset.l10nId = @pageTitle
+        if @pageTitleArgs
+          console.log(@pageTitleArgs, JSON.stringify(@pageTitleArgs))
```
philschatz: Is this an oversight?

src/scripts/modules/media/body/body.coffee
```diff
@@ -8,7 +8,7 @@ define (require) ->
   EditableView = require('cs!helpers/backbone/views/editable')
   ProcessingInstructionsModal = require('cs!./processing-instructions/modals/processing-instructions')
   SimModal = require('cs!./embeddables/modals/sims/sims')
-  Coach = require('cs!./embeddables/coach')
+  #Coach = require('cs!./embeddables/coach')
```
philschatz: Why was this commented? I _think_ this code is still actively being used. /cc @pandafulmanda

src/index.html
```diff
@@ -12,6 +12,14 @@
       <link rel="stylesheet" type="text/css" href="/styles/noscript.css" />
     </noscript>
 
+    <meta name="defaultLanguage" content="en-US">
+    <meta name="availableLanguages" content="en-US, pl">
+    <link rel="localization" href="/locale/{locale}/dictionary.ftl">
+
+    <script src="/scripts/browser.js"></script>
```
philschatz: Is there a way to include these in the single JS file instead of having multiple dist files? It seems that transpiling the l20n code using babel could still be done as a prerequisite step to compiling the rest of the coffeescript code.

src/scripts/modules/search/advanced/advanced-template.html
```diff
@@ -66,39 +66,31 @@
...
-                    {{native}} ({{english}})
-                  {{~/is~}}
-                </option>
-              {{~/each~}}
-            {{~/if~}}
-          </select>
-        </div>
-      </label>
+    	    <option value="" data-l10n-id="advanced-search-language-default-any">Any</option>
+                {{~#if filtered~}}
```
philschatz: Should these be outdented to line up with the "Any" option above?

src/scripts/pages/about/contact/contact-template.html
```diff
@@ -1,30 +1,34 @@
...
     OpenStax CNX<br />
     IBC Plaza Houston<br />
     5615 Kirby Drive, Suite 350<br />
-    Houston, TX 77005
+    Houston, TX 77005<br />
+    USA
   <p>
   <p>
-    (<a href="https://www.google.com/maps/d/embed?mid=zQw3pXjaT2oI.kPScb61jzndk">
+    (<a href="https://www.google.com/maps/d/embed?mid=zQw3pXjaT2oI.kPScb61jzndk" data-l10n-id="about-contact-google-maps">
```
philschatz: Nice (updating contact information for polish users)!

src/scripts/pages/donate/form/form.coffee
```diff
@@ -396,10 +396,20 @@ define (require) ->
       @amount = parseFloat(queryString.amount?.replace(',', '')) or 10
       @uuid = queryString.uuid
       @type = queryString.type
+      # Make sure event is detached from previous instantiation.
+      document.removeEventListener 'DOMRetranslated', @sortCountreies
+      # Sort country list alphabetically according to current locale.
```
philschatz: Ooh, nice!

src/scripts/helpers/backbone/views/base.coffee
ludekarts: Yes you're right. That was something for debugging purposes.

src/index.html
ludekarts: I think so. It is probably good idea to have it all compiled together. I'll try to make it works with Grunt and building process.

ludekarts:
@philschatz to address your questions:

1. It is not necessary to left the text in Handlebars templates, but we decided to left them there in case l20n fails from some reasons. This is also a nice way to have default fallback to English, and it makes internationalization less intrusive for the original code, so for example when editing template you know what content it will produce. Sometimes is hard to figured this out base only on the attributes keys.

2. I haven't used Travis before, so don't know if that is possible? Ideally you would want to install l20n with the rest of the dependencies through bower or npm. The only thing keeping us from doing that is lack of support for the language negotiation. We have sent the PR for this, but no response so far. We strongly believe that it will be pulled into the codebase since there was some initial implementation for this in previous versions.

3. No problem. Do you want to see there the full content of [IGFD](https://docs.google.com/document/d/19EY3YFrcVdbXC0KMFK8QCSqvYRXg_TtvbaNcUiDwXUI/edit?usp=sharing), or rather something more compressed?

One additional question from our side. Do you want, us to push changes for current issues to *I18n* branch, or we should wait for further reviews and more comments?

src/locale/pl/dictionary.ftl
```diff
@@ -0,0 +1,1950 @@
...
+
+advanced-search-sort-default-relevance = trafności
+
+advanced-search-sort-publication-date = Roku publikacji
+
+advanced-search-sort-popularity = popularności
+
+advanced-search-submit = Szukaj
+
+advanced-search-language-option = { $native } ({ $code ->
```
edwoodward: This summer, we had an intern modify this dropdown to display the actual languages we have content for rather than this list of all possibilities. Since this list is coming from the database now, how will that change your implementation?

philschatz:
@ludekarts Oops, I missed your question at the end of the comment; please do push changes to this branch 😄 . 

Also, it seems there is a conflict in `src/scripts/pages/donate/form/form.coffee`. Can you resolve the conflict? You can probably do `git rebase master` and then `git push --force-with-lease` just to make sure no one else pushed changes to this branch in the meantime.

philschatz:
It seems this PR now has merge conflicts; can you rebase `master` in to here?

src/scripts/config.js
```diff
@@ -160,8 +164,14 @@
 
       // concept-coach
       OpenStaxConceptCoach: {
-        deps: ['css!../../bower_components/concept-coach/main.min'],
+        deps: ['css!../../bower_components/concept-coach/assets/main'],
```
philschatz: This change seems to be causing me to not be able to run `grunt` or `grunt dist`. Why was it changed? Reverting it to the old code seems to fix the problem for me.

Log output:

```
$ grunt dist

Running "requirejs:compile" (requirejs) task
>> Error: ENOENT: no such file or directory, open
>> '/path/to/webview-katalyst/dist/scripts/../../bower_components/
>> concept-coach/assets/main.css'
>> In module tree:
>>     css/css
Warning: RequireJS failed. Use --force to continue.

Aborted due to warnings.
$
```

A similar error occurs due to the change in line 33 of this file

philschatz:
[[CTEK.01.01] Provide consultation for Katalyst internationalization work](https://trello.com/c/wUpXXehA/503-ctek-01-01-provide-consultation-for-katalyst-internationalization-work)

philschatz:
Are there a few simple tests that could be added to make sure the l20n code is running properly? (I remembered because Travis caught an error)

ludekarts:
Yes. I'll try to put some l20n tests together and send them with the next commit. I'll also try to fix the test that is failing now in Travis.

philschatz:
@ludekarts I _think_ I resolved conflicts in this PR but did not run it locally; I think the only changes were for the Summary toggle: `data-l10n-id="textbook-view-summary"`

bower.json
```diff
@@ -23,6 +23,8 @@
     "require-handlebars-plugin": "1.0.0",
     "aloha-editor": "oerpub/Aloha-Editor#master",
     "select2": "3.5.4",
+    "babel-polyfill": "^0.0.1",
+    "l20n": "https://github.com/tomekstach/l20n.js.git",
```
philschatz: @tomekstach : Could this forked repo be moved into the `openstax` or `katalysteducation` organization? GitHub supports redirects so you would not have to fix up this file (thank you for patiently rebasing all these times 😤  )

tomekstach: @philschatz : No problem, I will do it.

tomekstach: @philschatz : I transferred the `l20n.js` into the `katalysteducation` and also I changed this file, because I like the order :)

PR #1461, opened 2016-10-05
render the choices for multiple-choice questions, as well

reedstrm:
This is missing any  styling, or classes as hooks for styling, but works o.k.
All the questions on this page are acutally multiple choice:

http://cnx.org/contents/bDIuMp-w@1.1:XwiMLb76@2/Synthesis-of-Biological-Macrom

exercises - current
exercises - with choices

### DIFF
--- a/src/scripts/modules/media/body/body.coffee
+++ b/src/scripts/modules/media/body/body.coffee
@@ -257,7 +257,10 @@ define (require) ->
       embeddableItem.$el = @$el.find(embeddableItem.selector) if embeddableItem.selector
 
       $parent = embeddableItem.$el.parent()
-      embeddableItem.$el.replaceWith(embeddableItem.html)
+      if $parent.prop('tagName')?.toLowerCase() is 'p'
+        $parent.replaceWith(embeddableItem.html)
+      else
+        embeddableItem.$el.replaceWith(embeddableItem.html)
 
       embeddableItem.onRender?($parent)
 
--- a/src/scripts/modules/media/body/embeddables/exercise-template.html
+++ b/src/scripts/modules/media/body/embeddables/exercise-template.html
@@ -1,5 +1,14 @@
 {{#if data.items.[0].questions}}
     {{#each data.items.0.questions}}
-        {{{this.stem_html}}}
+        <div>{{{this.stem_html}}}</div>
+        {{#inArray this.formats  'multiple-choice'}}
+            {{#if this.answers}}
+            <ol>
+                {{#each this.answers}}
+                    <li>{{{this.content_html}}}</li>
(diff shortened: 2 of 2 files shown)


reedstrm:
@helenemccarron @zroehr Could use suggestions re: any structure or classes that might be needed to allow styling of these.

helenemccarron:
What are you trying to achieve exactly? These are EOC collations in the PDF and would be displayed as follow:

The class I found for it is `multiple-choice` and the answer is provided in the answer key.

reedstrm:
Ah, I suppose an example of the current output would help. The outside `<p>` is already in the source: the inserted content is the `div` and `ol`, that replace the `a` that pointed to the remote exercise.

```<p id="fs-id1167065844454">
        <div>Dehydration synthesis leads to the formation of what?</div>
            <ol>
                    <li>monomers</li>
                    <li>polymers</li>
                    <li>carbohydrates only</li>
                    <li>water only</li>
            </ol>
</p>```

**Comment by reedstrm** [comment]
FYI, the original version of this exercise, before it was extracted:
```html
<div data-type="problem" class="problem" id="fs-id2590162"><p id="fs-id886306">Dehydration synthesis leads to formation of</p>
<ol id="fs-id1238542" data-number-style="lower-alpha">
<li>monomers</li>
<li>polymers</li>
<li>water and polymers</li>
<li>none of the above</li>
</ol>
</div>
```

reedstrm:
New output:
```html
<div data-type="exercise" id="fs-id1167065763403" class="exercise os-exercise"><section>
<div data-type="problem" class="problem" id="fs-id1167065848055">
        <div>Dehydration synthesis leads to the formation of what?</div>
            <ol>
                    <li>monomers</li>
                    <li>polymers</li>
                    <li>carbohydrates only</li>
                    <li>water only</li>
            </ol>

</div>
</section></div>
```

philschatz:
Would it help if the entire `<div data-type='exercise'>` is replaced instead of just the `<div data-type="problem">`? I _think_ that was the original intention of the links (keep the `<exercise>` element so we know the link is an exercises, also bc CNXML does not allow block-level `<a>` tags)

reedstrm:
I think there are two issues here: dealing with the existing markup (which is an `a` in a `p` in a `problem div` in an `exercise div`) and exactly what we 'should' have. What I did replicates (for this one problem at least) what I see in the original book (Biology) modulo a `p`->`div` change. Since we'll have to deal with that structure anyway (when we do that book).

reedstrm:
bump. I think this is a reasonable fix for the current state. It restores the display to what you see in the equivalent content that has not had its exercises extracted.

PR #1467, opened 2016-10-31
coach needs a hot-fix for true-false questions

pandafulmanda:
coach update with [true-false fix](https://github.com/openstax/tutor-js/releases/tag/coach-20161031.a)


pandafulmanda:
close in favor of #1468, which is based off of the released version of webivew

PR #1496, opened 2017-01-23
Modified webmaster email

edwoodward:


edwoodward:
[BUG: Change contact email address from cnx@cnx.org to support@openstax.org](https://trello.com/c/r6XKsx41/21-bug-change-contact-email-address-from-cnx-cnx-org-to-support-openstax-org)

PR #1497, opened 2017-01-26
Feature flags and coach split

pandafulmanda:
related to openstax/tutor-js#1580 and openstax/tutor-deployment#272 and Connexions/cnx-deploy#195

Most of this is to take the changes being rebased between [production and master](https://github.com/Connexions/webview/compare/production) and pulling the differences into a `settings.js` flag.  This would me, no more rebasing!!!!

The other piece of this is to pull out `concept-coach` from the `bower` dependencies so that it loads from a remote `tutor` server.  There is some logic involved as well to ensure the freshest version of coach be loaded.  So, no more extra commits and webview deploys to update coach!

_Note_: while pulling session/authoring specific code into a separate script was explored, the file size difference in the end would be less than 3 kb so the extra weirdness and effort didn't seem worth it.

## Look ma!

Authoring and concept coach?!?!??!?!

screen shot 2017-01-30 at 4 31 57 pm


philschatz:
Cool! I like how you encapsulated most of the feature flag code into a `feature` module (except for handlebars templates it seems).

Just a couple questions:

1. How can I test it locally?
1. How do the deployment scripts toggle the flag? (via a config.js?)
1. When/how is Aloha loaded? At a cursory glance I do not see how it is being built/included

pandafulmanda:
@philschatz 

1. you can add `authoring` to the `features` in setting, or remove it
2. Connexions/cnx-deploy#195
3. The [aloha grunt task](https://github.com/Connexions/webview/blob/4e54ca53202a82d503f3357daf163b9b72dc1aa0/Gruntfile.coffee#L320-L329) and this [string-replace task](https://github.com/Connexions/webview/blob/4e54ca53202a82d503f3357daf163b9b72dc1aa0/Gruntfile.coffee#L269-L278) handle the differences between loading aloha locally vs when deployed

philschatz:
@RoyEJohnson can you verify that the authoring features only start up when the authoring flag is set? I'm having trouble starting it locally & knowing how to login to get to all the screens; you probably know better where all of those places are 😄

src/scripts/settings/config.js
```diff
@@ -0,0 +1,159 @@
+(function () {
```
edwoodward: Should this file be named something like cc-config.js since it seems to be all Concept Coach related. Just to make it clear to future devs and interns.

pandafulmanda: the intent of this is if there is necessary configuration in the future based on features, that this would be where the configuration would go.  but since it is just coach specific right now, should it still be changes to `cc-config.js`?

pandafulmanda: or perhaps, the coach specific bits can be pulled out into a separate file, leaving just the configuration logic in this file

Gruntfile.coffee
```diff
@@ -144,28 +144,28 @@ module.exports = (grunt) ->
...
+              'cs!pages/workspace/workspace'
+              'cs!pages/role-acceptance/role-acceptance'
               'cs!modules/media/editbar/editbar'
-              'cs!helpers/backbone/views/editable'
-
               'cs!configs/aloha'
             ]
             exclude: ['coffee-script', 'less/normalize']
-            excludeShallow: ['settings']
+            excludeShallow: ['settings', 'OpenStaxConceptCoach']
```
edwoodward: Just so I understand, this PR will only load Concept Coach when it is needed, correct? So the home page will not be downloading Concept Coach code and css.

pandafulmanda: yes

src/scripts/settings/base.js
```diff
@@ -0,0 +1,89 @@
...
+        'concepts-of-biology': 'b3c1e1d2-839c-42b0-a314-e119a8aafbdd@8.39',
+        'anatomy-and-physiology': '14fb4ad7-39a1-4eee-ab6e-3ef2482e3e22@6.19',
+        'introductory-statistics': '30189442-6998-4686-ac05-ed152b91b9de@17.20'
+      },
+
+      defaultLicense: {
+        code: 'by'
+      },
+
+      conceptCoach: {
```
edwoodward: Can this go into config.js?

pandafulmanda: no, these are settings for `conceptCoach` that can differ per environment

src/scripts/settings/config.js
edwoodward: If it is feature related maybe feature-config.js. I'm thinking new devs could be confused by settings.js and config.js. They are sort of the same thing from the file name. Maybe a comment in config.js explaining what it is for would be fine.

Gruntfile.coffee
edwoodward: 👍

src/scripts/settings/base.js
edwoodward: Makes sense now that I understand what config.js is for

src/scripts/settings/config.js
pandafulmanda: d885cc7

pandafulmanda: `config.js` renamed to `feature-config.js` in c4ad88b

RoyEJohnson:
@philschatz I KNOW NOTHING!

pandafulmanda:
@philschatz @RoyEJohnson, @DarkPrinceFrost and @pumazi  have been working with me to get [https://dev00.cnx.org](https://dev00.cnx.org/contents/v5a_xecj@3.4:-Yx1m5Xz@8/Themes-and-Concepts-of-Biology) working.  the authoring server-side isnt working completely on this server, but this is where we're at!

screen shot 2017-01-30 at 4 31 57 pm

pandafulmanda:
@edwoodward and @philschatz , i've pushed a couple more updates, and @pumazi  and @DarkPrinceFrost and I have tested it's deployment [on dev cnx](https://github.com/Connexions/webview/pull/1497#issuecomment-276214091)

edwoodward:
Second pass 👍

pandafulmanda:
No longer needed, `production` branch has been removed, editor no longer an actively used feature, `concept-coach` being deprecated out of webview in #1603

PR #1503, opened 2017-02-20
update coach to hotfix without sourcemap bloat

pandafulmanda:


pandafulmanda:
dont need to have matching update for `beta`.

PR #1504, opened 2017-02-20
update coach to hotfix without source map bloat

pandafulmanda:
a hotfix that will be deployed as [v0.21.14](https://github.com/Connexions/webview/releases/tag/v0.21.14)


pandafulmanda:
don't need to merge, hotfix has been deployed

PR #1505, opened 2017-02-20
Check subcol references

philschatz:
It seems the references to `id == 'subcol'` in the code are for editing-mode and are basically no-ops when just viewing content. But I added a check just-in-case 😄 

/cc @pandafulmanda since she was working on editing mode

### DIFF
--- a/src/scripts/models/content.coffee
+++ b/src/scripts/models/content.coffee
@@ -80,10 +80,13 @@ define (require) ->
         @set('childChanged', false)
 
     toJSON: (options = {}) ->
+      # This method is used for editing drafts as well as the JSON being sent to the handlebars template.
+      # Deleting the subcollection id may be problematic when attempting to use it in the
+      # Handlebars template but it does not seem to cause problems now.
       results = super(arguments...)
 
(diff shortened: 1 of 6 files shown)


src/scripts/models/content.coffee
```diff
@@ -80,10 +80,13 @@ define (require) ->
...
 
     toJSON: (options = {}) ->
+      # This method is used for editing drafts as well as the JSON being sent to the handlebars template.
+      # Deleting the subcollection id may be problematic when attempting to use it in the
+      # Handlebars template but it does not seem to cause problems now.
       results = super(arguments...)
 
       # Prevent new books with no id from being labelled a subcollection
-      if results.id is 'subcol'
+      if results.id is 'subcol' or results.collated
```
reedstrm: Hmm, this will always delete the subcol id if the viewed book is collated? My understanding of part of the 'delete the subcol id' was because it was a bad id: it repeated.

pandafulmanda: i think i will defer @reedstrm and you @philschatz on this...

PR #1508, opened 2017-02-24
update to coach build without global lodash

pandafulmanda:
and require underscore in body and coach just in case

[coach-20170223.b](https://github.com/openstax/tutor-js/releases/tag/coach-20170223.b)


philschatz:
Shucks, I would have thought that a linter would have caught the global reference to underscore...

pandafulmanda:
@philschatz it's from lodash and its kinda a weird bug

PR #1521, opened 2017-04-07
check that all strings are localized

philschatz:
From katalysteducation/webview#1 (see for some discussion).

This parses all of the HTML files (handlebars templates) and checks to see if any text nodes exist after doing the following:

1. removing the text nodes inside elements with `[data-l10n-id]`
1. remove `<script>` tags

I will try to circle back later (but no guarantees) to add better messages and maybe check that the `[data-l10n-id]` values exist

# To see it working...

add `<div>foo</div>` to the bottom of a `src/scripts/pages/*/*-template.html` file and run `npm run-script posttest`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,4 @@
 language: node_js
-node_js:
-  - '0.12'
-  - '6.9.1'
 before_install: ./script/bootstrap
 script: ./script/test
 sudo: false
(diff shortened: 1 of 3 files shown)


philschatz:
jsdom does not support node `0.12` and we have a `.nvmrc` file that uses `v6.9` so I just removed 0.12 from `.travis.yml`. 

Hopefully no objections 😄

RoyEJohnson:
Is this branch ready to delete?

philschatz:
oh, GitHub must've reopened the branch when I submitted the other PR (with the same branch name). I deleted the branch

PR #1524, opened 2017-04-10
update mathjax and move the CDN domain

philschatz:
From https://www.mathjax.org/cdn-shutting-down/ . This updates MathJax to 2.7.0 (from `2.6-latest`) and changes the domain to be cloudflare .

An alternative would be to build mathjax locally and host it as part of webview but that would increase the JS file size significantly. 

Chunking could bring the size back down a bit but would require a bit more dev work to only load MathJax when needed. And make sure the font files are hosted properly (correct path and content type headers)

### DIFF
--- a/src/scripts/config.js
+++ b/src/scripts/config.js
@@ -25,7 +25,7 @@
       'backbone-associations': '../../bower_components/backbone-associations/backbone-associations',
 
       // ## MathJax
-      mathjax: '//cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0/MathJax.js?config=MML_HTMLorMML',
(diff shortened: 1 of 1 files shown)


philschatz:
https://github.com/openstax/tutor-server/pull/1376 uses CDNJS so we should consolidate

reedstrm:
So we're updating version as well? Larger testing burden, but I guess …

philschatz:
oooh, `https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.6.0/MathJax.js` works so don't need to actually update the MathJax version 😄 

/cc @nathanstitt

reedstrm:
as does https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.6.1/MathJax.js

reedstrm:
Though I'd leave that up to QA - do we expect rendering diffs, and can we test for them now?

PR #1526, opened 2017-04-24
WIP Don't pass title into ftl template, might contain markup

reedstrm:
Baking files now adds markup to most page titles (we include the section number and spacer in spans). This is escaped when passed via `data-l10n-args` to the ftl translation templates, and ends up displayed as escaped markup, as seen here: 
screenshot from 2017-04-24 15-56-24

This patch converts some translations to not pass the page title, at the possible expense if less flexible translations. This is a WIP, since I have found more cases of title passed where I have not changed it. In general, if there's a single translated word or phrase followed by the content title, I'd suggest we instead of passing the title through l10n, just add a level of span around the translated bit.

Same page, with this patch:
screenshot from 2017-04-24 15-56-49

Fixes Connexions/cnx-rulesets#171

### DIFF
--- a/src/locale/en-US/dictionary.ftl
+++ b/src/locale/en-US/dictionary.ftl
@@ -595,7 +595,9 @@ textbook-view-btn-create = Create an Editable Copy
 
 textbook-view-publishing = publishing { $title }
 
-textbook-view-derived-from = Derived from <a href="{ $url }">{ $title }</a> by <span class="collection-authors">{ TAKE(50, $authors) }</span>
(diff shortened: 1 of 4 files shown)


reedstrm:
Thanks for the feedback - I knew this wasn't the correct way, but couldn't find how to turn off escaping. Sort of glad to know I didn't miss it in the docs.

philschatz:
Fixes Connexions/cnx-rulesets#171

reedstrm:
HAndled in a different way.

PR #1527, opened 2017-05-02
dont need the init since the documentReady pr has been merged

pandafulmanda:
https://github.com/katalysteducation/l20n.js/pull/1 merged, this is no longer needed, and may be causing problems sometimes on qa.


reedstrm:
looks correct on its face to me, and works from my local dev. Note that I can't repo the original bug locally, either.

PR #1529, opened 2017-05-17
Fix the bug about orgnization's name in the footer

tomjlw:
The footer of CNX lists 20 Million Minds Foundation, but that organization is now called the Michelson 20MM Foundation. Therefore I corrected it by replacing the old name with the new one.

### DIFF
--- a/src/scripts/modules/footer/footer-template.html
+++ b/src/scripts/modules/footer/footer-template.html
@@ -8,7 +8,7 @@
       </ul>
     </div>
     <div data-l10n-id="all-footer-support">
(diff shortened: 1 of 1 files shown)


reedstrm:
This fix did not work: the displayed text is actually coming from the translations over in src/locale/*/dictionary.ftl. Please correct there. @tomjlw

philschatz:
gaah, Should we just remove the text in the templates? That way it's clear that the `dictionary.ftl` file needs to be updated?

reedstrm:
Yes, then every time l20n breaks (it's sooo stable) our entire website will disappear!

PR #1532, opened 2017-05-18
strip HTML tags from the page title

philschatz:
As a stopgap solution while waiting for https://github.com/projectfluent/fluent.js/issues/37 to be resolved this strips the HTML tags out of the `<title>` and the Page Title.

/cc @reedstrm

### DIFF
--- a/src/scripts/helpers/backbone/views/base.coffee
+++ b/src/scripts/helpers/backbone/views/base.coffee
@@ -5,6 +5,7 @@ define (require) ->
   settings = require('settings')
   linksHelper = require('cs!helpers/links.coffee')
   locale = require('cs!helpers/locale.coffee')
+  stripTags = require('cs!../../strip-tags.coffee')
 
   dispose = (obj) ->
     delete obj.parent
@@ -77,12 +78,12 @@ define (require) ->
         historyPage = linksHelper.getCurrentPathComponents().page
 
       if @pageTitle
-        pageTitle = @pageTitle.replace(/<\/?[^>]+(>|$)/g, '') # Strip tags for baked HTML titles
-        document.title = pageTitle + settings.titleSuffix
+        strippedPageTitle = stripTags(@pageTitle)
+        document.title = strippedPageTitle + settings.titleSuffix
 
         if not isBook
           title = document.querySelector('title')
-          title.dataset.l10nId = pageTitle
+          title.dataset.l10nId = strippedPageTitle
 
           if @pageTitleArgs
             title.dataset.l10nArgs = JSON.stringify(_.extend({
--- a/src/scripts/helpers/backbone/views/base.coffee
+++ b/src/scripts/helpers/backbone/views/base.coffee
@@ -77,11 +77,12 @@ define (require) ->
         historyPage = linksHelper.getCurrentPathComponents().page
 
       if @pageTitle
-        document.title = @pageTitle + settings.titleSuffix
+        pageTitle = @pageTitle.replace(/<\/?[^>]+(>|$)/g, '') # Strip tags for baked HTML titles
+        document.title = pageTitle + settings.titleSuffix
 
         if not isBook
           title = document.querySelector('title')
-          title.dataset.l10nId = @pageTitle
+          title.dataset.l10nId = pageTitle
 
           if @pageTitleArgs
             title.dataset.l10nArgs = JSON.stringify(_.extend({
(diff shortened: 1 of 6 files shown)


src/scripts/helpers/backbone/views/base.coffee
```diff
@@ -77,11 +77,12 @@ define (require) ->
         historyPage = linksHelper.getCurrentPathComponents().page
 
       if @pageTitle
-        document.title = @pageTitle + settings.titleSuffix
+        pageTitle = @pageTitle.replace(/<\/?[^>]+(>|$)/g, '') # Strip tags for baked HTML titles
+        document.title = pageTitle + settings.titleSuffix
```
reedstrm: Will this (setting `document.title`) fix the  title-of-webpage issue? <head><title> equiv?

src/scripts/modules/media/header/header-template.html
```diff
@@ -14,7 +14,7 @@
     {{/if}}
 
     <div class="title">
-      <h2 data-l10n-id="textbook-view-header-{{#isnt status 'publishing'}}no-{{/isnt}}publishing{{#if chapter}}-chapter{{/if}}" data-l10n-args='{"chapter":"{{chapter}}","title":"{{safeQuotes pageTitle}}"}'>
+      <h2 data-l10n-id="textbook-view-header-{{#isnt status 'publishing'}}no-{{/isnt}}publishing{{#if chapter}}-chapter{{/if}}" data-l10n-args='{"chapter":"{{chapter}}","title":"{{stripTags pageTitle}}"}'>
```
reedstrm: For this specific one, I have yet to find a place where we render the page with a status if `publishing`, so let's strip it out completely.

philschatz:
Searching for a Page title in a non-baked book now looks correct:

image

src/scripts/helpers/backbone/views/base.coffee
philschatz: yeppers ; )

philschatz:
[Trello: BUG titles with HTML are escaped and displayed in places they should not be](https://trello.com/c/xYM422RX)

src/scripts/helpers/backbone/views/base.coffee
RoyEJohnson: A common way to strip html tags is to create a scratch div element, set its innerHTML to the HTML you want to strip, then take its `textContent`.

reedstrm: I like Roy's suggestion over the regex approach.

reedstrm:
BTW< this is exposed on production by doing an in-book-search that matches the title of a page. That page will then have escaped angle brackets added to it.

src/scripts/helpers/backbone/views/base.coffee
philschatz: The only downside I can think of is that this might cause untrusted code (`<script>alert("Hacked!")</script>`) to be executed in the browser if that is in the title (not 100% sure on this)

reedstrm:
ok the lack of highlighting for baked is odd - seems the highlight-search-results code needs a tweak. I'm willing to leave that till the next PR, given that production does not yet have baked books.

src/scripts/helpers/backbone/views/base.coffee
reedstrm: We have to trust the title coming from our DB: that's half the point of this. So no additional risk. But really, the browser executes scripts in an un-anchored div that is not part of the document?

RoyEJohnson: No, a script tag will only be executed if it appears in the document.

src/scripts/modules/media/header/header-template.html
```diff
@@ -14,15 +14,12 @@
     {{/if}}
 
     <div class="title">
-      <h2 data-l10n-id="textbook-view-header-{{#isnt status 'publishing'}}no-{{/isnt}}publishing{{#if chapter}}-chapter{{/if}}" data-l10n-args='{"chapter":"{{chapter}}","title":"{{safeQuotes pageTitle}}"}'>
-        {{#is status 'publishing'}}
-          <span class="label label-info">publishing</span>
-        {{/is}}
+      <h2>
         {{#if chapter}}<span class="title-chapter">{{chapter}}</span>{{/if}}
         {{{pageTitle}}}
```
philschatz: I don't know why the Page title is not highlighted... I switched this code to not go through l20n so it _should_ output highlighted...

philschatz:
@reedstrm I switched the code to use the "scratch div" method @RoyEJohnson suggested. It seems like the titles not highlighting in baked books is a BE issue. I commented on the line of code where the raw HTML from the search results is injected into the DOM

PR #1533, opened 2017-05-19
Fix the empty button issue

tomjlw:
When navigating to a button, descriptive text didn't present to screen reader users to indicate the function of the button. Therefore I fixed it by adding a descriptive text.

### DIFF
--- a/src/scripts/modules/media/body/body.coffee
+++ b/src/scripts/modules/media/body/body.coffee
@@ -157,7 +157,7 @@ define (require) ->
           .wrapInner('<section class="ui-body">')
           .prepend('''
             <div class="ui-toggle-wrapper">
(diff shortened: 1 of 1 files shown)


src/scripts/modules/media/body/body.coffee
```diff
@@ -157,7 +157,7 @@ define (require) ->
           .wrapInner('<section class="ui-body">')
           .prepend('''
             <div class="ui-toggle-wrapper">
-              <button class="btn-link ui-toggle" title="Show/Hide Solution"></button>
+              <button class="btn-link ui-toggle" title="Show/Hide Solution" aria-label="show/hide solution"></button>
```
edwoodward: How does this convey to the screen reader that this section is expandable or that the text changes when toggled. The aria-label is a good idea, but the solution should be in the CSS where the text on the button is changed.

PR #1536, opened 2017-05-22
Follow Up with the Pull Request  #1534

tomjlw:
Pull request doesn't pass the github test because Line 160 is too long. Therefore I split it into two lines.


pumazi:
Any reason you can't put this commit on #1534, since it still looks to be open?

PR #1539, opened 2017-05-26
Fix the Empty Button bug

tomjlw:
When navigating to a button, descriptive text didn't present to screen reader users to indicate the function of the button.
Therefore I fixed this bug by setting the aria-label according to the status of the attribute aria-expanded.

### DIFF
--- a/src/scripts/modules/media/body/body.coffee
+++ b/src/scripts/modules/media/body/body.coffee
@@ -469,7 +469,12 @@ define (require) ->
     toggleSolution: (e) ->
       $solution = $(e.currentTarget).closest('.solution, [data-type="solution"]')
       $solution.toggleClass('ui-solution-visible')
-
+      if $solution.hasClass('ui-solution-visible')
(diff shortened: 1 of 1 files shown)


src/scripts/modules/media/body/body.coffee
```diff
@@ -469,7 +469,12 @@ define (require) ->
     toggleSolution: (e) ->
       $solution = $(e.currentTarget).closest('.solution, [data-type="solution"]')
       $solution.toggleClass('ui-solution-visible')
-
+      if $solution.hasClass('ui-solution-visible')
```
philschatz: Can you remove this line and update the CSS to use `[aria-expanded="true"]` instead of `.ui-solution-visible`?

tomjlw: The first change is ok. But the problem is that if I follow the second change, the button will only be able to be cliked once. After that, however I clicked the button, the solution will neither display nor hide. I don't know why such a weired thing will happen.

PR #1544, opened 2017-06-07
Add additional navigation levels for the current table of contens

tomjlw:
Currently the Table of contents on webview shows chapters, and the pages in those chapters. Most of those pages are quite long, with logical sections inside that are linkable, and potentially useful for assignments, etc.


philschatz:
It seems like #1545 contains these commits plus some additional ones. Closing in favor of #1545

philschatz:
reopening so that the changes can be discussed separately (the comments found in #1545 still apply to this PR)

PR #1545, opened 2017-06-07
Link problem

tomjlw:
Users want to link to paragraphs in Rewrite. The ids exist, but the URL gets rewritten after the pages loads and the id gets lost.

image

### DIFF
--- a/src/scripts/modules/media/body/body.coffee
+++ b/src/scripts/modules/media/body/body.coffee
@@ -152,6 +152,31 @@ define (require) ->
             # NOTE: `.toggleClass()` explicitly requires a `false` (not falsy) 2nd argument
             $el.toggleClass('ui-has-child-title', $title.length > 0)
 
+          # Create subtitles in each chapter with corresponding ids as their href
+          total_subtitle = $temp.find("[data-depth='1']")
+          total_id = total_subtitle.children('[data-type="title"]')
+          id_length = total_id.length
+          i = 0
+          slash_index = window.location.href.lastIndexOf("/")+1
+          real_href = window.location.pathname
+          if real_href.indexOf("#")!= -1
+            hash_index = real_href.indexOf("#")-1
+            real_href = real_href.substring(0,hash_index)
+          location = $('.table-of-contents [href$="'+real_href+'"]').children(".title")
+          location2 = $('.table-of-contents [href$="'+real_href+'"]').children(".chapter-number")
+          # Keep adding subtitles until all subtitles are added
+          if location.children("li").length < id_length
+            while i < id_length
+              c_index = i + 1
+              id = total_subtitle[i].id
+              hash_id = "#" + id
+              list = $("<li>")
+              list.text(location2[0].textContent + "."  + c_index + " " + total_id[i].textContent)
+              list.wrapInner('''<a href = ""></a>''')
+              list.children("a").attr("href", hash_id)
+              location.append(list)
+              i = i + 1
+
           # Wrap solutions in a div so "Show/Hide Solutions" work
           $temp.find('.exercise .solution, [data-type="exercise"] [data-type="solution"]')
           .wrapInner('<section class="ui-body">')
@@ -469,6 +494,12 @@ define (require) ->
     toggleSolution: (e) ->
       $solution = $(e.currentTarget).closest('.solution, [data-type="solution"]')
       $solution.toggleClass('ui-solution-visible')
+      if $solution.hasClass('ui-solution-visible')
+        $solution.attr('aria-expanded',true)
+        $solution.attr('aria-label',"hide solution")
+      else
+        $solution.attr('aria-expanded',false)
+        $solution.attr('aria-label',"show solution")
 
     onEditable: () -> @$el.find('.media-body').addClass('draft')
 
--- a/src/scripts/helpers/links.coffee
+++ b/src/scripts/helpers/links.coffee
@@ -78,6 +78,8 @@ define (require) ->
     getCurrentPathComponents: () ->
       components = Backbone.history.fragment.match(@contentsLinkRegEx) or []
       path = components[0]
+      hash_path = window.location.hash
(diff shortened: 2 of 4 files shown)


philschatz:
Thanks for this! Can you edit the Pull Request body (click the ✏️  icon up above to the left of **Reviewers**) and include a screenshot of what the mini Table of Contents looks like?

src/scripts/modules/media/body/body.coffee
```diff
@@ -152,6 +152,31 @@ define (require) ->
             # NOTE: `.toggleClass()` explicitly requires a `false` (not falsy) 2nd argument
             $el.toggleClass('ui-has-child-title', $title.length > 0)
 
+          # Create subtitles in each chapter with corresponding ids as their href
```
philschatz: Does this method work when a Page is viewed outside of the book context? (Just the Page UUID, no Book UUID in the URL)

src/scripts/modules/media/body/body.coffee
```diff
@@ -152,6 +152,31 @@ define (require) ->
...
+          location = $('.table-of-contents [href$="'+real_href+'"]').children(".title")
+          location2 = $('.table-of-contents [href$="'+real_href+'"]').children(".chapter-number")
+          # Keep adding subtitles until all subtitles are added
+          if location.children("li").length < id_length
+            while i < id_length
+              c_index = i + 1
+              id = total_subtitle[i].id
+              hash_id = "#" + id
+              list = $("<li>")
+              list.text(location2[0].textContent + "."  + c_index + " " + total_id[i].textContent)
```
philschatz: I think this number is incorrect. It seems to be `{CHAPTER}.{SECTION_IN_THIS_PAGE}` which is misleading because students are probably expecting the number to be `{CHAPTER}.{SECTION_AKA_PAGE}.{SECTION_IN_THIS_PAGE}`

Is this numbering necessary or would it make sense to just number the sections in order, like https://en.wikipedia.org/wiki/OpenStax_College for example?

If this is intended to be a temporary solution then please ignore the following... I think this may need to be internationalized. For example, the `.` separator character means something different in English than it does in other languages (like Polish). That may be one reason why Wikipedia just numbers the sections

/cc @edwoodward

philschatz:
@tomjlw does the screenshot you included correspond to this PullRequest or to #1544 ? Or should one of these PR's be closed?

It seems like #1544 contains a subset of the commits in this PR

src/scripts/modules/media/body/body.coffee
```diff
@@ -152,6 +152,31 @@ define (require) ->
             # NOTE: `.toggleClass()` explicitly requires a `false` (not falsy) 2nd argument
             $el.toggleClass('ui-has-child-title', $title.length > 0)
 
+          # Create subtitles in each chapter with corresponding ids as their href
+          total_subtitle = $temp.find("[data-depth='1']")
```
edwoodward: @philschatz I think the top level is sufficient. We can discuss in scrum today.

src/scripts/modules/media/body/body.coffee
```diff
@@ -152,6 +152,31 @@ define (require) ->
...
+          slash_index = window.location.href.lastIndexOf("/")+1
+          real_href = window.location.pathname
+          if real_href.indexOf("#")!= -1
+            hash_index = real_href.indexOf("#")-1
+            real_href = real_href.substring(0,hash_index)
+          location = $('.table-of-contents [href$="'+real_href+'"]').children(".title")
+          location2 = $('.table-of-contents [href$="'+real_href+'"]').children(".chapter-number")
+          # Keep adding subtitles until all subtitles are added
+          if location.children("li").length < id_length
+            while i < id_length
```
tomjlw: Hi, Phil I think this will actually cause a problem because if I click a chapter several times, it will create many repetitive lists.

philschatz: Oh, nice catch! In that case maybe the last line would need to be something like:

```coffee
$location.empty()
$location.append($items)
```

src/scripts/modules/media/body/body.coffee
philschatz: ok. In that case can you make the CSS selector more specific? Just in case there are other elements that have a `data-depth='1'`.

Something like `...find("[data-type='section'][data-depth='1']")`

src/scripts/modules/media/body/body.coffee
reedstrm: Good Q: should we provide this as a TOC for inidividual page views?

tomjlw: Well, I created a simple table of ocntents for the individual page view, which is basically a <ol>, Is that enough? Or do I need to make it almost same as the table of contents in normal page?

src/scripts/modules/media/body/body.coffee
```diff
@@ -152,6 +152,31 @@ define (require) ->
             # NOTE: `.toggleClass()` explicitly requires a `false` (not falsy) 2nd argument
             $el.toggleClass('ui-has-child-title', $title.length > 0)
 
+          # Create subtitles in each chapter with corresponding ids as their href
+          total_subtitle = $temp.find("[data-depth='1']")
+          total_id = total_subtitle.children('[data-type="title"]')
```
reedstrm: Ah, I was asked about the diff in names w/ and w/o $ - good to know what they're there for.

src/scripts/modules/media/body/body.coffee
reedstrm: :+1:  @philschatz suggestion

src/scripts/modules/media/body/body.coffee
philschatz: I think that the code should not error when a user goes to an individual page. For example, https://cnx.org/contents/bG-_rWXy@5/Introduction (there is no left hand sidebar)

PR #1551, opened 2017-06-09
Link fix

tomjlw:
Users want to link to paragraphs in Rewrite. The ids exist, but the URL gets rewritten after the pages loads and the id gets lost.

### DIFF
--- a/src/scripts/helpers/links.coffee
+++ b/src/scripts/helpers/links.coffee
@@ -78,6 +78,8 @@ define (require) ->
     getCurrentPathComponents: () ->
       components = Backbone.history.fragment.match(@contentsLinkRegEx) or []
       path = components[0]
+      hash_path = window.location.hash
+      components[6] = hash_path
       if path?.slice(-1) is '/'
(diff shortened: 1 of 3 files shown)


reedstrm:
Why is this yet-another-branch, instead of being new commits to link-problem?

tomjlw:
@reedstrm It seems that I have messed up with link-problem and another branch. When I commit it directly into the link-problem, it will include the commit for additional navigation level.

reedstrm:
This does fix the going to the id on a page, but does not handle the in a book case.

https://cnx.org/contents/havxkyvS@9.423:Gv1bKcKt@#fs-idp132288224
goes to (approximately) the right spot on the page, but rewrites to:
https://cnx.org/contents/havxkyvS@9.423:Gv1bKcKt@6/Chemistry-in-Context
instead of 
https://cnx.org/contents/havxkyvS@9.423:Gv1bKcKt@6/Chemistry-in-Context#fs-idp132288224

PR #1555, opened 2017-06-15
Updating javascript packages

sisiz:
updating packages to the most recent versions, except alpha/beta versions.

running all the unit tests and running webview locally worked with no other changes made.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -22,7 +22,7 @@
     "require-less": "^0.1.5",
     "require-handlebars-plugin": "^1.0.0",
     "aloha-editor": "oerpub/Aloha-Editor#master",
-    "select2": "^4.0.3",
+    "select2": "3.0.3",
     "babel-polyfill": "^0.0.1",
     "l20n": "https://github.com/katalysteducation/l20n.js.git",
     "concept-coach": "openstax/tutor-js#coach-20170308.c",
--- a/bower.json
+++ b/bower.json
@@ -9,31 +9,31 @@
     "tests"
   ],
   "dependencies": {
-    "jquery": "2.1.1",
-    "underscore": "1.8.3",
-    "backbone": "1.2.3",
-    "backbone-associations": "0.6.2",
-    "bootstrap": "3.3.5",
-    "font-awesome": "4.4.0",
-    "requirejs": "2.1.20",
-    "requirejs-text": "2.0.14",
+    "jquery": "^3.2.1",
+    "underscore": "^1.8.3",
+    "backbone": "^1.2.3",
+    "backbone-associations": "^0.6.2",
+    "bootstrap": "^3.3.5",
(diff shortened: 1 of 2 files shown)


sisiz:
I reverted the select2 to 3.5.4 even though 4.0.3 is the newest version for now so that it doesn't break, still trying to understand why it not being installed by bower and was breaking the Build_CNX_webview check.

sisiz:
Select2 is updated to it's most recent version. But then I noticed when updating this warning came up:
`aloha-editor#fad076dbbd depends on font-awesome#~4.4 which resolved to font-awesome#4.4.0`
So font-awesome is not at the most recent (4.7.0) but instead (4.4.0)

bower.json
```diff
@@ -9,31 +9,31 @@
...
-    "bootstrap": "3.3.5",
-    "font-awesome": "4.4.0",
-    "requirejs": "2.1.20",
-    "requirejs-text": "2.0.14",
+    "jquery": "^3.2.1",
+    "underscore": "^1.8.3",
+    "backbone": "^1.2.3",
+    "backbone-associations": "^0.6.2",
+    "bootstrap": "^3.3.5",
+    "font-awesome": "fontawesome#^4.4.0",
```
philschatz: Why is there a need to prefix this with `fontawesome#`?

bower.json
```diff
@@ -9,31 +9,31 @@
...
   "dependencies": {
-    "jquery": "2.1.1",
-    "underscore": "1.8.3",
-    "backbone": "1.2.3",
-    "backbone-associations": "0.6.2",
-    "bootstrap": "3.3.5",
-    "font-awesome": "4.4.0",
-    "requirejs": "2.1.20",
-    "requirejs-text": "2.0.14",
+    "jquery": "^3.2.1",
```
philschatz: We originally pinned the versions to a specific version because sometimes minor version changes caused deployments to fail.

I wonder if you could investigate using yarnpkg.org to handle pinning the versions and keep the `^` in here  (run `yarn` instead of `npm install`)

edwoodward: @philschatz This sounds like a separate task. Can we add a card to the board and get this merged?

bower.json
sisiz: Not really sure, it had the prefix when first pulled the code from master.

bower.json
sisiz: @philschatz Should I change it to have the exact versions it has updated to rather than the `^` ?

philschatz: @sisiz if moving to yarn is a separate task then I think the `^` should be removed so deployments are still consistent

bower.json
philschatz: hmm, I don't see that in https://github.com/Connexions/webview/blob/master/bower.json . Could you switch it to just `4.4.0`?

PR #1556, opened 2017-06-16
Make donation email address configurable

tomjlw:
This allows email in the donation submission form to be changed via `settings.js`.

The email address in the "TNE_Admin_Email" field needs to be changed to openstaxgiving@rice.edu as opposed to invoices@cnx.org.

Also, this is the fix for webview only, I will have another pull request in the cnx-deploy .


tomjlw:
@philschatz Yes, that's because invoices@cnx.org has already been changed to openstaxgiving@cnx.org. But according to Ed's requirement, I need to create a variable in setting.js and read the value of that variable and then pass the value into html. Then if we need to change the e-mail address in the future, we just need to change the varibale's value in setting.js.

philschatz:
Ah ok. I was wondering why `invoices@cnx.org` was in the Pull Request description. I added a line in the Pull Request Description and clarified the title of this Pull Request.

PR #1558, opened 2017-06-21
Addition toc

tomjlw:
Phil suggests me to use .map instead of while loop to iterate over the elements. However, I tried but it always shows the problem, therefore here I still use the while loop.

### DIFF
--- a/src/scripts/modules/media/body/body.coffee
+++ b/src/scripts/modules/media/body/body.coffee
@@ -152,6 +152,27 @@ define (require) ->
             # NOTE: `.toggleClass()` explicitly requires a `false` (not falsy) 2nd argument
             $el.toggleClass('ui-has-child-title', $title.length > 0)
 
+          # Create subtitles in each chapter with corresponding ids as their href
+          $totalSubtitle = $temp.find("[data-depth='1']")
+          $totalId = $totalSubtitle.children('[data-type="title"]')
+          idLength = $totalId.length
+          countIndex = 0
+          realHref = window.location.pathname
+          #Determine if this is a outer page view or not
+          $location = $(".table-of-contents [href$='#{realHref}'] > .title")
+          # Keep adding subtitles until all subtitles are added
+          if $location.children("li").length < idLength
+            while countIndex < idLength
+              subIndex = countIndex + 1
+              subId = $totalSubtitle[countIndex].id
+              hashId = "#" + subId
+              $list = $("<li>")
+              $list.html('&nbsp;'+'&nbsp;'+'&nbsp;'+'&nbsp;'+'&nbsp;'+'&nbsp;'+$totalId[countIndex].textContent)
+              $list.wrapInner('''<a href = ""></a>''')
+              $list.children("a").attr("href", hashId)
+              $location.append($list)
+              countIndex += 1
+
           # Wrap solutions in a div so "Show/Hide Solutions" work
           $temp.find('.exercise .solution, [data-type="exercise"] [data-type="solution"]')
           .wrapInner('<section class="ui-body">')
--- a/src/scripts/modules/media/tabs/contents/toc/page.coffee
+++ b/src/scripts/modules/media/tabs/contents/toc/page.coffee
@@ -53,6 +53,11 @@ define (require) ->
       e.stopPropagation()
 
       $link = $(e.currentTarget)
+      if $link[0].href.indexOf("#") > -1
+        index = $link[0].href.indexOf("#")
+        new_link =  $link[0].href.substring(index)
+        window.location.hash = new_link.substring(1)
+        return
       router.navigate $link.attr('href'), {trigger: false}, () => @trackNav()
       @model.get('book').setPage($link.data('page'))
 


src/scripts/modules/media/tabs/contents/toc/page.coffee
```diff
@@ -53,6 +53,11 @@ define (require) ->
       e.stopPropagation()
 
       $link = $(e.currentTarget)
+      if $link[0].href.indexOf("#") > -1
```
philschatz: Rather than using `$link[0].href` can you pull the value into a local variable? See https://github.com/Connexions/webview/pull/1545#discussion_r120736703

src/scripts/modules/media/body/body.coffee
```diff
@@ -152,6 +152,27 @@ define (require) ->
...
+          #Determine if this is a outer page view or not
+          $location = $(".table-of-contents [href$='#{realHref}'] > .title")
+          # Keep adding subtitles until all subtitles are added
+          if $location.children("li").length < idLength
+            while countIndex < idLength
+              subIndex = countIndex + 1
+              subId = $totalSubtitle[countIndex].id
+              hashId = "#" + subId
+              $list = $("<li>")
+              $list.html('&nbsp;'+'&nbsp;'+'&nbsp;'+'&nbsp;'+'&nbsp;'+'&nbsp;'+$totalId[countIndex].textContent)
```
philschatz: Why is `&nbsp;` used here? If it is for aligning the text, it might be cleaner to do this in CSS.

src/scripts/modules/media/body/body.coffee
```diff
@@ -152,6 +152,27 @@ define (require) ->
...
             $el.toggleClass('ui-has-child-title', $title.length > 0)
 
+          # Create subtitles in each chapter with corresponding ids as their href
+          $totalSubtitle = $temp.find("[data-depth='1']")
+          $totalId = $totalSubtitle.children('[data-type="title"]')
+          idLength = $totalId.length
+          countIndex = 0
+          realHref = window.location.pathname
+          #Determine if this is a outer page view or not
+          $location = $(".table-of-contents [href$='#{realHref}'] > .title")
```
philschatz: Was the conclusion from #1545 that these elements were going to be added to the Table of Contents (Tree) Model and use the existing Tree UI rather than recreating the UI from within here?

src/scripts/modules/media/body/body.coffee
```diff
@@ -152,6 +152,27 @@ define (require) ->
             # NOTE: `.toggleClass()` explicitly requires a `false` (not falsy) 2nd argument
             $el.toggleClass('ui-has-child-title', $title.length > 0)
 
+          # Create subtitles in each chapter with corresponding ids as their href
+          $totalSubtitle = $temp.find("[data-depth='1']")
```
philschatz: See https://github.com/Connexions/webview/pull/1545#discussion_r120947015

src/scripts/modules/media/body/body.coffee
```diff
@@ -152,6 +152,27 @@ define (require) ->
...
+          $totalId = $totalSubtitle.children('[data-type="title"]')
+          idLength = $totalId.length
+          countIndex = 0
+          realHref = window.location.pathname
+          #Determine if this is a outer page view or not
+          $location = $(".table-of-contents [href$='#{realHref}'] > .title")
+          # Keep adding subtitles until all subtitles are added
+          if $location.children("li").length < idLength
+            while countIndex < idLength
```
philschatz: I'm curious, why did `.map(...)` not work? See https://github.com/Connexions/webview/pull/1545#discussion_r120741070

src/scripts/modules/media/body/body.coffee
tomjlw: After the discussion between Ed and Ross, the final solution will be just change the display instead of changing the structure of TOC

src/scripts/modules/media/body/body.coffee
tomjlw: Here I use &nbsp; to make indent for subtitles.

src/scripts/modules/media/body/body.coffee
tomjlw: Well, the console says "cannot read property 'ownerDocument' of undefined" when I applied the following codes:
 $items = $temp.find('section[data-depth="1"]').map (index, node) ->
            # Get the section id so we can link to it
            id = $(node).attr('id')
            # Get a copy of the section title to add to the link
            $title = $(node).find('> [data-type="title"]').clone()
            # Build the list item and link
            $listItem = $("<li/>")
            $link = $("<a href='##{id}'/>")
            $link.append($title)
            $listItem.append($link)
            # Return the list item
            $listItem
$location.append($items)

tomjlw: And later I changed the map function provided by you a little bit , and now it works.
`a=$temp.find('section[data-depth="1"]')
          $items = $.map(a, (node) ->
          # Get the section id so we can link to it
            id = $(node).attr('id')
            # Get a copy of the section title to add to the link
            $title = $(node).find('> [data-type="title"]').clone()
            # Build the list item and link
            $listItem = $("<li/>")
            $link = $("<a href='##{id}'/>")
            $link.append($title)
            $listItem.append($link)
            # Return the list item
            $listItem
         )`

kerwinso:
See https://github.com/Connexions/webview/pull/1566

PR #1559, opened 2017-06-23
fix note titles, make example & note styling consistent for baked books

zroehr:
fix [#206](https://github.com/Connexions/cnx-recipes/issues/206)
Also, make note styling consistent across all note and example/exercise types

### DIFF
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -14,10 +14,17 @@ h6 {
 }
 
 ul, ol {
-  margin-top: 1rem;
-  margin-bottom: 0;
   color: @gray;
+  margin-bottom: 1rem;
+  margin-top: 1rem;
 }
+
+iframe {
+  display: block;
+  margin: 3rem auto;
+  width: 100%;
+}
+
 
 // It is already displayed at the top of the page in blue
 // the `>` is because collated titles (like Key Concepts) contain descendent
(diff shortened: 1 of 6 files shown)


kerwinso:
QA checklist:

- [ ]  Econ (https://github.com/Connexions/cnx-recipes/issues/206)
- [ ] University Physics (https://github.com/Connexions/cnx-recipes/issues/272)

philschatz:
Related to https://github.com/Connexions/cnx-recipes/issues/271

philschatz:
@zroehr Can you include some screenshots of the new code in the Pull Request description above? That makes it easier to review the changes.

src/scripts/modules/media/body/content-baked.less
```diff
@@ -122,36 +143,83 @@ h1.example-title .text {
 
 // Helper mixin for expanding all the selectors
 .make-block(@type) {
+  #blockish>.style(@type);
```
philschatz: It looks like this is copy/pasta'd from somewhere else. Can the code be organized so that this is written in one place and called in 2 places?

zroehr: @philschatz whenever I remove L145-L186 because `.make-block(@type)` is already called in content-common, it does not work. for some reason, the styles in content-baked for notes aren't overriding those in content-common unless I define `.make-block(@type)` again

philschatz: hmm, that's odd. Could https://github.com/Connexions/webview/blob/f7d43d9d0cc8c1d7ec76c60e2df972a5653a15e8/src/scripts/modules/media/body/content-common.less#L580-L583 be moved into both `content-baked.less` and `content-raw.less`? 

Then `content-common.less` would only define `make-block(...)` and it would only be called in either `-baked.less` or `-raw.less`?

zroehr: @philschatz that is not working either. the styles under `#blockish` in `content-baked.less` aren't being overridden.

PR #1560, opened 2017-06-27
fix embedded video iframe width

zroehr:
Fixes https://github.com/Connexions/cnx-recipes/issues/208


zroehr:
will fix in PR #1559

PR #1562, opened 2017-06-29
fix history issue, fix anchor link position - beware wall of text

caderitter:
## Problem description
 if a user follows an inter-book link in baked content (i.e. Review Questions to Answer Key), the URL being navigated away from is not stored in browser history, yet two extra entries are stored in history, one being the long href with a hash (i.e. `#fs-id2197874-solution`), and one without a hash. Finally, the URL is canonicalized and updated, and stored in history. 

Example URL: 
```
/contents/185cbf87-c72e-48f5-b51e-f14f21b5eabd@10.61:a2d22c3d-7bba-5629-bb1a-a2e5eb375e0f@10.61#fs-id2197874-solution
```

## Expected behavior
User follows link, origin URL is stored, two intermediate URLs are not stored, final URL is stored. 

## Bottom line for my changes
Back button now works reliably, but still stores extra URLs in the history. ~~Navbar hiding anchors is fixed.~~

## What I need help with :)

This problem ended up being decently complicated. Backbone's `navigate` is extended in `history.coffee` such that it removes the regular functionality of `navigate`: normally, when `{trigger: true, replace: true}` is passed in to `navigate`, the route is triggered but the URL is **not** updated, [as seen in the docs](http://backbonejs.org/#Router-navigate). However, from `history.coffee`:
```
@history[if options.replace then 'replaceState' else 'pushState']({}, document.title, url)
```
As far as I know, this is equivalent to `history.replaceState()` which actually replaces the most recent entry in the browser history with a given URL. An erroneous call to this was replacing the URL being navigated away from (one part of the problem).

That's fine - what sucks is, I need _exactly_ what the functionality was before (so we can navigate to a route without updating the URL). I don't want to change the way `navigate` is extended because I'm afraid it will break all the other usages in the application. However, I think that's what will have to happen. **_Does anyone have any input on this?_**

Right now, the back button functionality is fixed in that it is usable (the user can actually get back to the original page), but you still must click back 3 times to get to the page you clicked the link on. 

Interestingly, if [these lines](https://github.com/Connexions/webview/blob/0c7b32b9da36c89d57c0854348457d9226c68ff7/src/scripts/modules/media/body/body.coffee#L449-L451) which handle jumping to a hash are removed, the two "long" URLs are **not** cached in browser history. Of course, this breaks jumping to hashes.  **_Any ideas about that?_**

~~Sorry for wall of text, but one more thing - I added a one-line change at [line 455](https://github.com/Connexions/webview/blob/0c7b32b9da36c89d57c0854348457d9226c68ff7/src/scripts/modules/media/body/body.coffee#L455) which scrolls up by the height of the navbar, so anchors aren't covered.~~

### DIFF
--- a/src/scripts/modules/media/body/body.coffee
+++ b/src/scripts/modules/media/body/body.coffee
@@ -449,8 +449,7 @@ define (require) ->
           @fragmentReloaded = true
 
           hash = window.location.hash
-          window.location.hash = ''
-          window.location.hash = hash
+          window.location.replace(hash)
 
       $target = $(window.location.hash)
       if $target.prop('tagName')?.toLowerCase() is 'iframe'
--- a/src/scripts/modules/media/body/body.coffee
+++ b/src/scripts/modules/media/body/body.coffee
@@ -452,10 +452,6 @@ define (require) ->
           window.location.hash = ''
           window.location.hash = hash
 
-          # scroll up by width of nav bar
-          # FIXME - make this dynamic
-          window.scrollBy(0, -145)
-
       $target = $(window.location.hash)
       if $target.prop('tagName')?.toLowerCase() is 'iframe'
         $target.on('load', jumpToHash)
--- a/src/scripts/modules/media/body/body.coffee
+++ b/src/scripts/modules/media/body/body.coffee
@@ -140,9 +140,9 @@ define (require) ->
                     [data-type="example"], [data-type="exercise"], [data-type="note"]').each (index, el) ->
             $el = $(el)
             $contents = $el.contents().filter (i, node) ->
-              return !$(node).is('.title, [data-type="title"]')
+              return !$(node).is('.title, [data-type="title"], .os-title')
             $contents.wrapAll('<section>')
-            $title = $el.children('.title, [data-type="title"]')
+            $title = $el.children('.title, [data-type="title"], .os-title')
             $title.wrap('<header>')
             # Add an attribute for the parents' `data-label`
             # since CSS does not support `parent(attr(data-label))`.
--- a/src/scripts/modules/media/body/body.coffee
+++ b/src/scripts/modules/media/body/body.coffee
@@ -136,8 +136,9 @@ define (require) ->
           $temp.children('[data-type="abstract"]').remove()
 
           # Wrap title and content elements in header and section elements, respectively
-          $temp.find('.example, .exercise, .note,
-                    [data-type="example"], [data-type="exercise"], [data-type="note"]').each (index, el) ->
+          $temp.find('.example, .exercise, .note, .abstract,
+                    [data-type="example"], [data-type="exercise"],
+                    [data-type="note"], [data-type="abstract"]').each (index, el) ->
             $el = $(el)
             $contents = $el.contents().filter (i, node) ->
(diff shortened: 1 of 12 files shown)


src/scripts/modules/media/body/body.coffee
```diff
@@ -439,21 +441,26 @@ define (require) ->
...
       jumpToHash = () =>
         if currentPage.get('loaded') and not @fragmentReloaded and window.location.hash
           @fragmentReloaded = true
           hash = window.location.hash
           window.location.hash = ''
           window.location.hash = hash
 
+          # scroll up by width of nav bar
+          # FIXME - make this dynamic
+          window.scrollBy(0, -145)
```
philschatz: Thanks for trying to fix this problem but I think this might break responsiveness. 

There was some discussion on how to handle this in general and I think the decision was to have the top nav bar work the same was [news](https://www.washingtonpost.com/news/posteverything/wp/2017/06/28/the-supreme-court-is-in-no-hurry-to-protect-voters-from-gerrymandering/#article-body) websites and GitHub do it (by hiding the nav bar when you scroll down and re-show it when you scroll up)

Can you remove this from this Pull Request?

caderitter: makes sense. will do!

philschatz:
@caderitter is there an Issue related to this PR?

caderitter:
@philschatz no, but there's a [trello card](https://trello.com/c/ihhMXAOO/68-bug-new-baked-pages-are-not-being-added-to-browser-history)

philschatz:
Assigned to @kerwinso for testing 😄

kerwinso:
Cade said:
> Right now, the back button functionality is fixed in that it is usable (the user can actually get back to the original page), but you still must click back 3 times to get to the page you clicked the link on.

This updated behavior was confirmed on [University Physics on qa.cnx.org](https://qa.cnx.org/contents/1Q9uMg_a@5.22:cqPvIeML@5.22/Index):
1. Go to index of book
1. Click term
1. scroll up on the ensuing page slightly to see the term (because the top nav is hiding it)
1. click the Back button in the browser 3 times to get back to the Index.

It's an improvement, but still not expected/ standard behavior for the browser Back button.

kerwinso:
image

Stackblocks:
We shouldn't be expecting a user to repeat an action in order to achieve a different result on a specific repetition. This is a good start but still needs to have some work done. As it stands I don't believe this is a roadblock for AW1, so no worries there.

src/scripts/modules/media/body/body.coffee
zroehr: Im working on implementing the updated header functionality stated by @philschatz in PR https://github.com/Connexions/webview/pull/1541

caderitter:
By the way, all the nastiness in "files changed" is due to my rebasing of my branch with master

kerwinso:
Tested on qa.cnx.org with Biology and University Physics, on Chrome 59 incognito, OS X El Capitan. 

The Back button now works as expected, including within a book (between EOC and Answer Key, Index reference and term, etc.) Clicking it once brings you back to where you expect. Hurray!

PR #1566, opened 2017-07-05
Addition toc

tomjlw:
Need to create additional Table of contents

### DIFF
--- a/src/scripts/modules/media/media.coffee
+++ b/src/scripts/modules/media/media.coffee
@@ -94,10 +94,23 @@ define (require) ->
       mainPage.regions.main.append(footerNav)
       footerNav.$el.addClass('footer-nav')
       @mainContent = windowWithSidebar.regions.main
-
       $pinnable = @regions.pinnable.$el
       pinnableTop = $pinnable.offset().top
       $toc = tocView.$el
+      
+      tocModel = tocView.model
+      @listenTo(tocModel, 'add:contents change:contents', ->
+        mediaBodyView.on('render', ->
+          $totalSubtitle = $("body").find("section [data-depth=1]")
+          subTable=$.map($totalSubtitle, (node) ->
+            subTitles=$(node).find('> [data-type="title"]').clone().text()
+            subIds="#" + $(node).attr("id")
+            return {text:subTitles,id:subIds}
+          )
+          tocModel.get('contents').models[1].attributes.contents.models[1].set({subTable: subTable})
+        )
+      )
+
       isPinned = false
       setTocHeight = ->
         tocTop = $pinnable.height()
@@ -176,7 +189,6 @@ define (require) ->
       mediaBodyView.on('render', ->
         scrollTo = if wasPinnedAtChange then pinnableTop + 1 else 0
         $(window).scrollTop(scrollTo)
-        tocView.generateAdditionTOC()
       )
 
     updateSummary: () ->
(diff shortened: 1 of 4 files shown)


src/scripts/modules/media/tabs/contents/toc/page-template.html
```diff
@@ -5,6 +5,11 @@
         {{#if chapter}}<span class="chapter-number">{{chapter}}</span>{{/if}}
         <span class="title{{#if active}} active{{/if}}" data-depth="{{depth}}">
           {{{title}}}
+          {{#if active}}
+          {{#each subTable}}
+            <li><a href={{this.id}}>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;{{this.text}}</a></li>
```
philschatz: Can you update the CSS to indent this sublist rather than using `&nbsp;`

src/scripts/modules/media/media.coffee
```diff
@@ -94,10 +94,23 @@ define (require) ->
...
+      tocModel = tocView.model
+      @listenTo(tocModel, 'add:contents change:contents', ->
+        mediaBodyView.on('render', ->
+          $totalSubtitle = $("body").find("section [data-depth=1]")
+          subTable=$.map($totalSubtitle, (node) ->
+            subTitles=$(node).find('> [data-type="title"]').clone().text()
+            subIds="#" + $(node).attr("id")
+            return {text:subTitles,id:subIds}
+          )
+          tocModel.get('contents').models[1].attributes.contents.models[1].set({subTable: subTable})
```
philschatz: Why is the second item in `.models[]` updated instead of the first item? Can you add a JavaScript comment on this line so when someone else looks over this code in the future they will know what is happening?

Also, does this code cause an error when a person views a Page that is not in a book?

src/scripts/modules/media/tabs/contents/toc/page.coffee
```diff
@@ -53,6 +53,11 @@ define (require) ->
       e.stopPropagation()
 
       $link = $(e.currentTarget)
+      if $link[0].href.indexOf("#") > -1
+        index = $link[0].href.indexOf("#")
+        new_link =  $link[0].href.substring(index)
+        window.location.hash = new_link.substring(1)
```
philschatz: I wonder if this is impacted by the work @caderitter is doing

src/scripts/modules/media/media.coffee
tomjlw: Actually, I am wrong with these two subscripts. These two subscripts should correspond to indexes of current model and current page. I made some changes and added some comments.
And According to Ed and Ross's requirement, I don't need to add additional TOC for page that is not in a book. After my change, there is no error, just nothing happens because such kind of page doesn't have table of contents.

philschatz:
I was unable to see the ToC update. Am I doing something wrong?

I did the following:

1. check out the `additionTOC` branch
2. run `npm start`
3. visit http://localhost:8000
4. click "Biology"

cnx-toc-not-populating-sub-toc

tomjlw:
@philschatz Could you try it in Non-biology book like Sociology? It seems biology, concepts of biology and Anatomy & Physiology(basically the book which contains Unit besides chapter) have a different model structure

philschatz:
Thanks. It works for Sociology but I am still unable to get it to work with Biology.

Also, it seems the links in the ToC are missing some CSS styling. See:

image


I believe they should be green like the other links in the ToC

edwoodward:
@philschatz Is there anything to salvage here. The intern that worked on this is no longer here.

PR #1568, opened 2017-07-06
bold chapter number in TOC and page title

zroehr:


src/scripts/modules/media/header/header.less
```diff
@@ -30,6 +30,10 @@
       padding-right: 0.4em;
     }
 
+    .os-number {
+      font-weight: 500;
```
philschatz: Why use `500` here and `bold` in `section.less`?

zroehr: The chapter number in the header appeared too bold when using `bold`

kerwinso:
Passed QA. Rebaked University Physics, now the chapter numbers in the Contents and page title are bold face:
image

PR #1573, opened 2017-07-11
Preload fix

tomjlw:
Use Prerender to reduce the loading time


kerwinso:
Trello card: https://trello.com/c/Llzbaxc8/160-investigate-using-preload-on-cnxorg

kerwinso:
It is indeed slightly faster.

PR #1574, opened 2017-07-13
fix summary toggle

zroehr:
move data-is-baked to main-content div
hide abstract in media-header when data-is-baked is true


kerwinso:
(temporarily) fixes https://github.com/Connexions/cnx-recipes/issues/205 and https://github.com/Connexions/cnx-recipes/issues/214

kerwinso:
Passed dev testing. The summary toggle for baked (Openstax) books is now hidden:

image

For non-baked (non-Openstax) books (such as [Understanding Basic Music Theory](https://qa.cnx.org/contents/KtdLe6cv@3.74:fWmMtOTx@8/Rhythm)), the summary toggle still displays. This is the desired behavior:

image

PR #1577, opened 2017-08-02
fix reload twice bug

zroehr:
fix reload twice bug
fix page jump after hash change
fix index term cover up by header

# Screencast

Sorry it's slow because it is running in Dev mode and Phil's trackpad is very old

cnx-back-button

### DIFF
--- a/src/scripts/modules/media/body/body.coffee
+++ b/src/scripts/modules/media/body/body.coffee
@@ -70,7 +70,7 @@ define (require) ->
       canonicalPath = linksHelper.getPath('contents', {model: @model, page: pageId}, []) + window.location.hash
       if (canonicalPath isnt "/#{currentRoute}")
         # replace previous URL with the canonical path
-        history.replaceState({}, @model.get('title'), canonicalPath)
+        router.navigate(canonicalPath, {replace: true})
 
     updateTeacher: ($temp = @$el) ->
       $els = $temp.find('.os-teacher')
@@ -83,7 +83,7 @@ define (require) ->
     goToPage: (pageNumber, href) ->
       @model.setPage(pageNumber)
 
-      router.navigate href, {trigger: true, replace: false}, => @parent.parent.parent.trackAnalytics()
+      router.navigate href, {replace: false}, => @parent.parent.parent.trackAnalytics()
 
     getCoach: ->
       moduleUUID = @model.getUuid()?.split('?')[0]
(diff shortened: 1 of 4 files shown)


philschatz:
Awesome! I didn't think the 3-sentence description did it justice so I added a screencast to the Pull Request Body above

cnx-back-button

src/scripts/modules/media/media.coffee
```diff
@@ -111,7 +114,7 @@ define (require) ->
       adjustHashTop = ->
         handleHeaderViewPinning()
         if isPinned
-          obscured = $pinnable.height()
+          obscured = $pinnable.height() + 50
```
philschatz: How well does this works when the window width is smaller (like on a mobile phone)?

If this number will continue to be used can it be moved into a common file so it is not duplicated?

zroehr:
Thanks Phil!

src/scripts/modules/media/media.coffee
```diff
@@ -67,6 +67,11 @@ define (require) ->
 
     triggerHashChange: (e) ->
       Backbone.trigger('window:hashChange')
+      href = $(e.currentTarget).attr('href')
+
+      e.preventDefault()
+      $elHash = href.split("#")[1]
```
philschatz: Is this variable used anywhere?

src/scripts/modules/media/body/body.coffee
```diff
@@ -445,11 +445,11 @@ define (require) ->
...
       # to force the browser window to find the intended content (as a side effect)
       jumpToHash = () =>
-        if currentPage.get('loaded') and not @fragmentReloaded and window.location.hash
-          @fragmentReloaded = true
-
+        if currentPage.get('loaded') and window.location.hash
           hash = window.location.hash
-          window.location.replace(hash)
+          $pinnable = $('.pinnable')
+          obscured = $pinnable.height() + 50
```
philschatz: How well does this works when the window width is smaller (like on a mobile phone)?

If this number will continue to be used can it be moved into a common file so it is not duplicated?

caderitter:
So happy this got fixed 😄  thanks so much for jumping on this!!

src/scripts/modules/media/media.coffee
zroehr: good catch, this was left over from some code I removed and is not used.

src/scripts/modules/media/body/body.coffee
zroehr: I will verify this works on small screens and move it into a common file

src/scripts/modules/media/media.coffee
zroehr: I will verify this works on small screens and move it into a common file

philschatz:
The link scroll seems to not move the target item to the top of the page. See:

cnx-link-scroll

Is that the desired behavior?

zroehr:
I added some offset so the link wouldn't sit right at the bottom of the header after a click. This offset also allows for feature boxes to be shown in full when a jump link is linking to a note feature. I can adjust this if needed.

zroehr:
@philschatz heres an example of what im referring to when it comes to notes

kerwinso:
Dev testing failed. See https://github.com/Connexions/cnx-recipes/issues/213 and https://github.com/Connexions/webview/issues/1579. Also, a minor regression found in https://github.com/Connexions/webview/issues/1578.

kerwinso:
Passed QA:

- Hash links go where they're supposed to (including footnote references) without reloading the page. 
- Back button in the browser works, even when navigating between 2 hash links. 
- Index terms and other content arrived at via hash link are no longer hidden by the top nav header. 

Such a better user experience!

PR #1576, opened 2017-08-01
Footerfix

tomjlw:
Fix some words in the attribution tag according to the requirement.

PR #1581, opened 2017-08-09
change email address in footer to supportATopenstax.org

kerwinso:
Instead of all the changes in https://github.com/Connexions/webview/pull/1576, Alina requested we just change the email address in the footer for assignable webview.


kerwinso:
Verified on qa.cnx.org:
image

PR #1583, opened 2017-08-30
add developer/debugging tools

philschatz:
[\[Trello Story\]](https://trello.com/c/iuwxbWMP/213-webview-developer-debug-tools)

Inspired heavily by http://ks52.web.rice.edu/urlcommands.html , this adds several useful links into the `More Information` tab at the bottom of every page. There might be a better place to "enable" this information.

# Screenshots

This shows the really ugly list-o-links (at the bottom). Not pretty, but it works!

image

Enabling/disabling dev mode: This:

1. starts with dev mode disabled
1. shows how to enable it (or just add `?dev-mode=true` to the browser URL)
1. shows that it remains enabled when you navigate to other pages
1. shows that it disables when you click "Enabled"

webview-dev-mode-quick

# TODO

- [x] use the `settings.js` URLs for archive and legacy so this works in any environment
- [x] add a link-to-me hover on every element that has an id attribute (maybe only blockish ones)
    - this is so QA/content people can easily link to a specific element
- [x] render a border around content elements and show the class names on that element above the box (hopefully just using CSS)
- [ ] Add a sneaky `console.log` message so people on the internet know where to find our source code

/cc @kerwinso @pumazi and @karenc for feedback


kerwinso:
@philschatz This looks really cool! A couple of thoughts off the bat:

1) It looks like developer mode would be really easy for any front-end user to find. Yes our content is open, but my concern is that users would stumble onto dev mode and start randomly clicking links ("Enqueue PDF?  What does this do?"), which could impact server performance.

Rather than having a link on the front end, I think @pumazi had the idea to enable dev mode with a super secret keyboard shortcut. Would it be possible to implement this instead?

2) For the info box, instead of `id` can we call that field `uuid`?

philschatz:
I thought that operations like "Enqueue PDF" require authentication (and so it wouldn't impact server performance), no? but we could do something like what Android does: "Click on the Rice Logo 10 times and dev mode is enabled". I'm worried that a super-secret code would be too difficult to remember. Although.... I think it would be cool if it was https://en.wikipedia.org/wiki/Konami_Code

pumazi:
The secret key sequence code is already implemented in webview (at least the editing end of it) for PI (print instruction?) directives. This is one of the reasons why the key sequence was suggested.

kerwinso:
@philschatz The URL commands/ operations in question do not require authentication (shhh!). This includes downloading files (zips, PDFs), which could also impact server performance.

kerwinso:
@philschatz Does one of the links prepend `archive-` to the URL so you can see the HTML for that page? I think that may be the `Source: XHTML` link in your screenshot.

Also, cc'ing @reedstrm in case he knows of additional dev/debugging tools that are not represented here.

philschatz:
Yep, that's the link. Here are all of the URLs: https://github.com/Connexions/webview/pull/1583/files#diff-216a08dfc178e8e722d7727d70c663cfR152

Unfortunately you need to toggle between "Page" and "Book" to see all the links. I wonder if the debugging list should contain:

1. A link to the Book in  the archive
1. A link to the Page in the archive
1. A link to the Page in the Book in the archive

Currently it only does the first 2

reedstrm:
Looks pretty good - I'd definitely add table/page archive links for all three flavors. Also, no reason to make the dev links change w/ the tab-slider - put all the book/page links (all three flavors) on one page.

As to access, yes this is a bit too tempting - and no, enqueue is _not_ protected right now. Either have it be a url_param that sets a persistent cookie, then have the "exit dev mode" change the cookie so you get the link to turn it back on. If you dump cookies (or use a different browser),  you need to add the url param again.

edwoodward:
@philschatz Please make the suggested changes or close this PR.

philschatz:
@edwoodward Addressed by only exposing "Dev Mode" when it is enabled (you have to know how to turn it on so users will not stumble upon it).

src/styles/main.less
```diff
@@ -256,3 +256,14 @@ p:empty {
...
   border-top: 4px solid !important;
 }
+
+[data-dev='control-box'] {
+  background-color: #f99;
+  position: fixed;
+  top: 50px;
+  left: 10px;
+}
+// This does not seem to work for some reason. Maybe XHTML mode?
```
edwoodward: Do we need to keep the commented out code?

src/scripts/modules/media/body/content-common.less
```diff
@@ -711,3 +711,28 @@ table {
     margin: 1rem 3rem;
   }
 }
+
+
+// Dev-mode styling
+* > [data-dev='anchor'] {
+  text-decoration: none !important;
+
+  // This method worked when it was just an icon, not when it contains the data-type
```
edwoodward: Should the commented out code remain in the file?

brittweinstein:
@philschatz can this be closed?

PR #1600, opened 2018-03-02
Don't use page number in links to other versions of book

reedstrm:
The 'this isn't the latest version click here" message uses an old url composed of bookid and page number. That doesn't survive crossing book versions w/ different numbers of pages (before the page in question). We're hitting this old bug more now, because going from unbaked -> baked will reliably add one or more pages to every single chapter.

### DIFF
--- a/src/scripts/helpers/links.coffee
+++ b/src/scripts/helpers/links.coffee
@@ -59,7 +59,7 @@ define (require) ->
 
       if model.isBook?()
         title = trim(model.get('currentPage')?.get('title'))
-        page = ":#{model.getPageNumber()}"
+        page = ":#{model.get('currentPage')?.getShortUuid()}"
 
       return "#{settings.root}contents/#{id}#{page}/#{title}"
 


reedstrm:
Yup, the steps to find this (feels kind of like hunting bug through zope/plone skins and archetypes):

1. grep source for "newer version", find https://github.com/Connexions/webview/blob/master/src/scripts/modules/media/latest/latest-template.html Note that it uses a variable `url`
2. Check https://github.com/Connexions/webview/blob/master/src/scripts/modules/media/latest/latest.coffee discover that url is defined in terms of `linksHelper.getModelPath(@model)` which is imported as `linksHelper = require('cs!helpers/links')`
3. Hunt down https://github.com/Connexions/webview/blob/master/src/scripts/helpers/links.coffee#L54-L65 and the above fix.
4. Then, grep and discover that is used in exactly 2 places: the "latest" module, and the "footer" "metadata" module, used for the "more information" tab at the bottom, but the module attribute set with it (model.url) seems to go unused. Even if it did, this is a more correct value :-)

PR #1602, opened 2018-03-06
add support for analytics array

philschatz:
This changes the analytics code to support an array of analytics IDs rather than a single analytics ID.

The various call sites may be good candidates for cleaning up when exactly we send an analytics event.

# Notes

One thing that might be worth including soon (maybe in a subsequent PR) is to fix when these analytics events are being fired. It might be worth spending time to write unit tests so they don't break in the future.

### DIFF
--- a/src/scripts/helpers/handlers/analytics.coffee
+++ b/src/scripts/helpers/handlers/analytics.coffee
@@ -14,18 +14,25 @@ define (require) ->
         l: Date.now()
 
     # Send the current page to every analytics service
-    send: (account, fragment = Backbone.history.fragment) ->
-      if not /^\//.test(fragment) then fragment = '/' + fragment
+    sendAnalytics: (accounts, fragment = Backbone.history.fragment) ->
+
+      # temporarily support a single field for analytics (sometimes it is `null`)
+      unless Array.isArray(accounts)
+        console.log 'Consider calling analytics.sendAnalytics with an Array'
+        @sendAnalytics([accounts], fragment)
 
       require ['analytics'], (ga) =>
-        # Use the default analytics ID in settings if no account is specified
-        account ?= settings.analyticsID
 
-        # TODO investigate if we need specific names for our tracker name
-        trackerName = @getTrackerName(account)
-        ga('create', account, 'auto', trackerName)
+        accounts.forEach (account) =>
+          if not /^\//.test(fragment) then fragment = '/' + fragment
+          # Use the default analytics ID in settings if no account is specified
+          account ?= settings.analyticsID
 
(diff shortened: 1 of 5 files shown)


reedstrm:
Perhaps I Was too quick to merge this.  While running tests against my backend modifications to serve multiple GA codes, I only get errors from this code, specifically `Uncaught TypeError: Cannot read property 'forEach' of undefined` Using the debugger, I do indeed find that this code (line 26 above) is only called with `accounts` `undefined`. Even though the codes are sent in the json, as expected. Hmmm. @philschatz ??

src/scripts/helpers/handlers/analytics.coffee
```diff
@@ -14,18 +14,25 @@ define (require) ->
...
-        # Use the default analytics ID in settings if no account is specified
-        account ?= settings.analyticsID
 
-        # TODO investigate if we need specific names for our tracker name
-        trackerName = @getTrackerName(account)
-        ga('create', account, 'auto', trackerName)
+        accounts.forEach (account) =>
+          if not /^\//.test(fragment) then fragment = '/' + fragment
+          # Use the default analytics ID in settings if no account is specified
+          account ?= settings.analyticsID
```
reedstrm: This is never reached if accounts above is empty, is it? I think it's not possible. So the site-wide code will never be sent. Tracing through currently-released production, I see this send method being called twice on each page navigation - once w/ "undefined" account, and once w/ the book-defined account code.

philschatz:
Oh yeah, this probably should not have been merged without checking against the backend changes.

https://github.com/Connexions/webview/blob/ac52dd639ae45fdb1d9cbb81cfedf1df4baedb2d/src/scripts/helpers/handlers/analytics.coffee#L16-L35 should probably be replaced with:

```coffee
   # Send the current page to every analytics service
    sendAnalytics: (accounts, fragment = Backbone.history.fragment) ->

      # Use the default analytics ID in settings if no account is specified
      unless accounts
        accounts = [settings.analyticsID]
        
      # temporarily support a single field for analytics (sometimes it is `null`)
      unless Array.isArray(accounts)
        console.log 'Consider calling analytics.sendAnalytics with an Array'
        @sendAnalytics([accounts], fragment)

      require ['analytics'], (ga) =>

        accounts.forEach (account) =>
          if not /^\//.test(fragment) then fragment = '/' + fragment

          # TODO investigate if we need specific names for our tracker name
          trackerName = @getTrackerName(account)
          ga('create', account, 'auto', trackerName)

          ga("#{trackerName}.send", 'pageview', fragment)
```

I will submit a PR

PR #1603, opened 2018-03-08
no more coach in webview

pandafulmanda:


kerwinso:
@pandafulmanda How would these changes be tested?

philschatz:
@kerwinso it looks like you might be able to check these pieces of content to verify that Concept Coach is no longer loading. Also, it would likely require testing all of the other viewing functionality on cnx.org to make sure it still works

kerwinso:
@philschatz Thanks, but I need to know where in the user-facing part of webview to check these changes. If it's not obvious or user-facing, but still needs QA, then I'd need a more specific test plan written by a developer. As far as testing the other viewing functionality on CNX, that will be covered by our existing regression tests.

reedstrm:
@philschatz got time to review? Otherwise, I'll merge.

reedstrm:
@kerwinso this code has already been shut off for production, by removing the list of CC books from the config. This change removes functionality only, but you can't actually see that, since we already removed it from the UI. Basically, I feel this is regression only at this point.

PR #1604, opened 2018-03-09
update canonical link for each page

reedstrm:
This will now correctly set the `<link rel="canonical">` header to the value of the book uuid + page uuid, without versions, rather than just the book uuid for all pages. This is needed stop confusing Google and hypothes.is as to the correct URL to use for a given page.

### DIFF
--- a/src/scripts/helpers/links.coffee
+++ b/src/scripts/helpers/links.coffee
@@ -51,7 +51,7 @@ define (require) ->
       return url
 
     # Get the URL to view a given content model
-    getModelPath: (model, withTitle = true) ->
+    getModelPath: (model, withTitle) ->
       page = ''
       id = model.getUuid?() or model.id
       version = model.get?('version') or model.version
--- a/src/scripts/helpers/links.coffee
+++ b/src/scripts/helpers/links.coffee
@@ -51,7 +51,7 @@ define (require) ->
       return url
 
     # Get the URL to view a given content model
-    getModelPath: (model) ->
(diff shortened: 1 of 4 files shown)


src/scripts/helpers/links.coffee
```diff
@@ -51,17 +51,20 @@ define (require) ->
...
+    getModelPath: (model, withTitle) ->
       page = ''
       id = model.getUuid?() or model.id
       version = model.get?('version') or model.version
       title = trim(model.get?('title') or model.title)
 
       if model.isBook?()
         title = trim(model.get('currentPage')?.get('title'))
-        page = ":#{model.get('currentPage')?.getShortUuid()}"
+        page = ":#{model.get('currentPage')?.getUuid()}"
```
philschatz: Why the switch from short UUID's? The browser contains short UUIDs... It seems like this would cause every search result URL to not match (or closely match) the URL that students/instructors will paste into their LMS (or bookmark)

reedstrm: Because the uuid is the canonical version: we 301 redirect shortIds to the long one - Also, it wasn't shortIds, it was a mixed url w/ book uuid + page shortId - that exists nowhere else.

reedstrm:
Why? We have other cases of a defaulted non-explicit arg, so it seems an arbitrary style choice that you're enforcing on my code. ;-) Also, by doing it my way, I explicitly don't have to touch the every place the function is used. Do we have a JS-lint standard anywhere, BTW? We have one for python.

PR #1606, opened 2018-03-20
Display for Books a page is in

sisiz:
creating a side nav bar for books a page is in when a page is navigated to outside the context of a book.
Displays the title of the book, revision date and author.

Notes:
book title cuts off after two lines. User just has to click on the book if they want to see full title (no openable twisty). This was fred's suggestion for UX

EDIT 
Books in Page button loading correctly and not displaying for book, only for pages.

# Screenshot

image


reedstrm:
So, I couldn't help myself, I was going to do one or two, then I did the whole thing. Only thing left is to get the Polish translations in place.

src/scripts/modules/media/tabs/contents/contents-template.html
```diff
@@ -10,6 +10,51 @@
     </div>
     <div class="result-count" data-l10n-id="textbook-editor-contents-results-matches" data-l10n-args='{"hits":{{resultCount}} }'>{{resultCountText}}</div>
   {{/if}}
-  <div class="toc">
-    {{! TocView}}
-  </div>
+
+  {{#if isBook}}
```
reedstrm: @philschatz apparently, on first load this isBook is False, while the one in nav is true. I.e. we see a 'Contents' button, but the TOC is not rendered.

reedstrm:
Hadn't realized the isBook() method was a local addition in this PR in `content.coffee`. This looks correct to me now.

reedstrm:
@philschatz this has been through several rounds of UX review - we have more vertical than horizontal, actually, so a single line would be suboptimal. Go argue w/ Fred :-) Merging.

PR #1608, opened 2018-03-29
do not redirect when unicode characters are in the URL

philschatz:
When a URL contains unicode characters, the page just constantly reloads. This code was originally added in #1319 to handle `@` and `:` but it might no longer be necessary


reedstrm:
But if I stop blinking, my eyes get all dry …

philschatz:
ok, I checked that all the links work in Chrome, FireFox, and Safari. Since this changes how all the URLs are processed, I also clicked About, Search, in-book-search, Next/Previous

PR #1619, opened 2018-04-02
simplified book list template - fixes line break on Edge after bullet

reedstrm:
This fixes the rendering on Edge, while maintaining the Chrome-only 2-line-with-ellipsis rendering. Firefox and Edge both show all three lines (of the exemplar page at /contents/xelFKBHs@6/Systems-Development-Life-Cycle )

### DIFF
--- a/src/scripts/modules/media/tabs/contents/contents-template.html
+++ b/src/scripts/modules/media/tabs/contents/contents-template.html
@@ -25,28 +25,23 @@
       <ul>
         {{#each booksIn}}
         <li >
-            <div class="max-wrap-two-lines">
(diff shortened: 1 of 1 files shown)


reedstrm:
@philschatz I stripped out a bunch of apparently extraneous `<div>` wrappers, and moved the `max-wrap-two-lines` class to the `<a>` itself.

reedstrm:
*sigh* this apparently does _not_ fix it in Edge - I suspect the combo of removing the `div` _as well as_ moving the class to the `a` still triggers the extra linebreak.

reedstrm:
Nope, doesn't work. `display: -webkit-box` is sufficient to break Edge

Issue #1620, opened 2018-04-03
UX: 'This page is in books' loading message in blank pane looks like a bug


kerwinso:
https://staging.cnx.org/contents/6AjDYAxi@16/Composizione-Grafica-in-Proces

![image](https://user-images.githubusercontent.com/11298165/38257269-887dde02-3725-11e8-9af8-48e5283958ca.png)

## wait for it...
![image](https://user-images.githubusercontent.com/11298165/38257289-9072a8a4-3725-11e8-864c-38ed18d4618e.png)

reedstrm:
@sisiz Here's the bug.

kerwinso:
https://trello.com/c/7XpChr94/436-ux-this-page-is-in-books-loading-message-in-blank-pane-looks-like-a-bug

kerwinso:
Looks good on qa server. The loading pane updates so fast I recorded a video/animated gif of what it looks like loading this page: https://qa.cnx.org/contents/yWshIYVW@1/How-Changes-in-the-Cost-of-Liv

Watch the left pane closely:

![1620](https://user-images.githubusercontent.com/11298165/38278160-84809b38-3760-11e8-8265-36cb5514429b.gif)

In any event, definitely an improved user experience.

kerwinso:
FYI, this is what it looks like on staging during those 3-4 seconds of loading time. The left pane is blank but open:

![image](https://user-images.githubusercontent.com/11298165/38322150-37013a44-37ff-11e8-8828-ac96531fc1a9.png)

cc @openstaxalina in case you want to consider a different user experience with Fred (e.g., default books list closed)

PR #1621, opened 2018-04-03
Wait to load book count

sisiz:
Waiting to load the text of the book count until the count has actually been loaded.
Also adding special case text for if there are 0 or 1 books found.
(example page in one book. not sure where to find one in 0 books to look at http://localhost:8000/contents/e6qfF73c@3/131-Chromosomal-Basis-of-Inher)

(https://github.com/Connexions/webview/issues/1620)

### DIFF
--- a/src/scripts/modules/media/nav/nav.coffee
+++ b/src/scripts/modules/media/nav/nav.coffee
@@ -174,3 +174,6 @@ define (require) ->
       @regions.tocPanel?.show(new ContentsView({model: @model}))
       @enableClearSearch() if @model.get('searchResults')?
       @updateToc()
+      if @model.isPage()
+        @toggleBooksList()
+        @closeContentsOnSmallScreen()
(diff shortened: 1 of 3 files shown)


reedstrm:
So, we need some code that watches the appropriate model, so that the booklist is opened once it's loaded.

sisiz:
sorry, what is the issue? When the data-l10n-* args replace the span with the translated text why would it not be correct?

reedstrm:
Just figured it out - the default text that was in the template would render _before_ the model loaded, that's where the "This page is in  books" flash was coming from. The simplest fix is to just not have any default text, depend on l20n to load the text from the dictionary.ftl file

reedstrm:
Tested locally, this _does_ also solve the problem, since the numBooksText resolves to [empty string] until extras is loaded. I think we'll merge this, since it gives us the code to consider other solutions in the future.

PR #1624, opened 2018-04-03
added aria-role to the button, and change appearance onfocus

sisiz:
Update the button for Get this book to have the tag `aria-role="dialogue"` because that is what the trello card says to do. Not really sure what that actually does with the screen reader.
Also update the button to display when the focus is on it by inverting the color (normally green with white text. when in focus white with green text)

code changes for this card (https://trello.com/c/j91OmKs2/434-a11y-get-this-book-button-should-have-aria-roledialog)

### DIFF
--- a/src/scripts/modules/media/header/header-template.html
+++ b/src/scripts/modules/media/header/header-template.html
@@ -35,9 +35,9 @@
       <div class="downloads">
         {{#or hasDownloads buyLink}}
           {{#is mediaType 'application/vnd.org.cnx.collection'}}
-            <button class="btn" tabindex="0" type="submit" data-l10n-id="textbook-view-btn-get-this-book">Get This Book!</button>
+            <button class="btn getContent" tabindex="0" type="submit" data-l10n-id="textbook-view-btn-get-this-book" aria-role="dialoge">Get This Book!</button>
           {{else}}
-            <button class="btn" tabindex="0" type="submit" data-l10n-id="textbook-view-btn-get-this-page">Get This Page!</button>
+            <button class="btn getContent" tabindex="0" type="submit" data-l10n-id="textbook-view-btn-get-this-page" aria-role="dialoge">Get This Page!</button>
           {{/is}}
         {{/or}}
       </div>
--- a/src/scripts/modules/media/header/header.less
+++ b/src/scripts/modules/media/header/header.less
@@ -8,6 +8,14 @@
   }
 }
 
+.getContent:focus {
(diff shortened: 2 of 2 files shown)


src/scripts/modules/media/header/header-template.html
```diff
@@ -35,9 +35,9 @@ <h2>
       <div class="downloads">
         {{#or hasDownloads buyLink}}
           {{#is mediaType 'application/vnd.org.cnx.collection'}}
-            <button class="btn" tabindex="0" type="submit" data-l10n-id="textbook-view-btn-get-this-book">Get This Book!</button>
+            <button class="btn getContent" tabindex="0" type="submit" data-l10n-id="textbook-view-btn-get-this-book" aria-role="dialoge">Get This Book!</button>
```
philschatz: In this repository the class names tend to be `hyphen-separated` rather than `camelCase`. Is there a specific reason for using `camelCase`?

src/scripts/modules/media/header/header-template.html
```diff
@@ -35,9 +35,9 @@ <h2>
...
           {{else}}
-            <button class="btn" tabindex="0" type="submit" data-l10n-id="textbook-view-btn-get-this-page">Get This Page!</button>
+            <button class="btn getContent" tabindex="0" type="submit" data-l10n-id="textbook-view-btn-get-this-page" aria-role="dialoge">Get This Page!</button>
```
philschatz: Should this attribute be `aria-role="dialog"` ?

philschatz:
I think the `role="dialog"` attribute is necessary in this case because there it no HTML element that describes a dialog.

I am not sure if a label and description are necessary but it seems like a good thing to include.

The value for a label and description can either be provided by:

- `aria-label="English label"` and `aria-description="English description"`
- or, `aria-labeledby="id-of-the-element-containing-the-label"` and `aria-describedby="id-of-the-description-element"`

src/scripts/modules/media/header/header-template.html
kerwinso: Also note the spelling of 'dialog'.

kerwinso:
@philschatz to discuss button onfocus changes with @sisiz.

kerwinso:
@philschatz I got webview running locally so I could take a look at the onfocus style change. This is what it looks like after you click the green "Get This Book!" button:

image

kerwinso:
Fixed with https://github.com/Connexions/webview/issues/1823

Issue #1609, opened 2018-03-29
Answer Key entries on new line after number


Stackblocks:
# **Description of the issue**
In the Answer Key, answers are appearing on the line below the associated numbers.

# **Screenshots**
![image](https://user-images.githubusercontent.com/12678235/38112360-4f49e502-3367-11e8-8c17-f5bf97789f02.png)

# **Impact**
-   [col23873](https://devb.cnx.org/contents/WUj6TfJ0@5.1:pOpsYFzJ@5.1/Answer-Key) Answer Key

Stackblocks:
Displaying correctly with new code.

![image](https://user-images.githubusercontent.com/12678235/38582540-b03a4e84-3cd5-11e8-8d7c-c5a4095126e7.png)

PR #1625, opened 2018-04-04
Answer Key entries on new line after number

brittany-johnson:
For issue #1609


brittany-johnson:
@philschatz created an issue so we won't forget https://github.com/Connexions/webview/issues/1627

PR #1628, opened 2018-04-12
Mobile scrolling

sisiz:
changing mobile scrolling so that when scrolling down the entire header goes away, and then when scrolling back up the nav bar (contents button and title) comes back but you still have to scroll all the way to the top for the full header. Unless the toc is open then the nav bar and toc remain pinned Regular view remains unchanged

### DIFF
--- a/src/scripts/modules/media/media.less
+++ b/src/scripts/modules/media/media.less
@@ -11,7 +11,7 @@
     width: 100%;
     max-width: @max-viewport-width;
     z-index: 1;
-    transition: top 1s ease-in-out;
+    transition: top 0.4s ease-in-out;
   }
 
   .pinnable.closed{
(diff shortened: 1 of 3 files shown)


src/scripts/modules/media/media.less
```diff
@@ -5,12 +5,23 @@
   height: 100%;
 
   .pinnable.pinned {
+    opacity: 1;
     position: fixed;
     top: 0;
     width: 100%;
     max-width: @max-viewport-width;
     z-index: 1;
+    transition: top 1s ease-in-out;
```
reedstrm: make this 0.1 sec and were good

sisiz:
Changed time to 0.4, 0.1 seemed a bit fast because of how big the header is. But still has a bug with being at the very top then scrolling down then up really fast. (The commit fix toc scroll mobile works for the header to just appear and leave. But the ease in/out part in the most recent commits has some bugs)

reedstrm:
Emily gave UX sign-off (with a bit of optional style tweak, that I managed to fit in :-) )

PR #1629, opened 2018-04-13
Add ids to navbar links to facilitate better automation

m1yag1:
* adds `id="nav-browse"` to the browse link
* adds `id="nav-about"` to the about link
* adds `id="nav-donate"` to the donate link

### DIFF
--- a/src/scripts/modules/header/header-template.html
+++ b/src/scripts/modules/header/header-template.html
@@ -39,9 +39,9 @@
       {{! Collect the nav links, forms, and other content for toggling }}
       <div class="collapse navbar-collapse" id="page-nav">
         <ul class="nav nav-justified">
-          <li {{#is page 'browse'}}class="active"{{/is}}><a href="/browse" data-l10n-id="all-navbar-search"></a></li>
-          <li {{#is page 'about'}}class="active"{{/is}}><a href="/about" data-l10n-id="all-navbar-about-us"></a></li>
-          <li {{#is page 'donate'}}class="active"{{/is}}><a href="/donate" data-l10n-id="all-navbar-give"></a></li>
+          <li class="nav-browse {{#is page 'browse'}}active{{/is}}"><a href="/browse" data-l10n-id="all-navbar-search"></a></li>
+          <li class="nav-about {{#is page 'about'}}active{{/is}}"><a href="/about" data-l10n-id="all-navbar-about-us"></a></li>
+          <li class="nav-donate {{#is page 'donate'}}active{{/is}}"><a href="/donate" data-l10n-id="all-navbar-give"></a></li>
           <li>
             <a href="http://www.rice.edu">
               <img src="/images/rice.png" class="rice" alt="Rice University" width="74" height="32" />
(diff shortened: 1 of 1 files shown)


kerwinso:
https://trello.com/c/UqElSYkz/471-added-classes-to-navbar-links-to-facilitate-better-automation

src/scripts/modules/header/header-template.html
```diff
@@ -39,9 +39,9 @@
       {{! Collect the nav links, forms, and other content for toggling }}
       <div class="collapse navbar-collapse" id="page-nav">
         <ul class="nav nav-justified">
-          <li {{#is page 'browse'}}class="active"{{/is}}><a href="/browse" data-l10n-id="all-navbar-search"></a></li>
-          <li {{#is page 'about'}}class="active"{{/is}}><a href="/about" data-l10n-id="all-navbar-about-us"></a></li>
-          <li {{#is page 'donate'}}class="active"{{/is}}><a href="/donate" data-l10n-id="all-navbar-give"></a></li>
+          <li class="navbar-browse{{#is page 'browse'}} active{{/is}}"><a href="/browse" data-l10n-id="all-navbar-search"></a></li>
```
philschatz: Could we use `id="..."` instead of a class attribute?
Class names have a couple of "unique" features compared to other attributes like `id="..."` and `data-foo=".."`:

- they are usually used for

philschatz: Also, since this attribute is only for automation, would it make sense to prefix the `id` with `automation-` or something so devs know it doesn't interfere with the app or any styling in the CSS?

m1yag1: It would be nice it it stayed more relevant to the element that we need to select. The locators look like the following:

```
_root_locator = (By.ID, 'header')
_nav_locator = (By.ID, 'page-nav')
_search_input_locator = (By.ID, 'find_content_input')
```
If this sort of change is too complicated i'll just use css selectors and nth children.

reedstrm: I appreciate the desire to make these tests less fragile. nth children is a red-flag for me for that sort of thing. As to the specific changes nav items should be unique-on-the-page, so `id`s sound right to me, as well.  The 'all-navbar-about-us'  thing is used for localization - its the key into the locale dictionaries. 

As to naming, having a policy that "every page structure element that causes navigation when clicked on should have an id starting with `nav-`" seems reasonable to me, without being to verbose. @philschatz how's that sound?

philschatz: Could just the link text be used (rather than creating an attribute or using `:nth`)? That would also help verify that the internationalization code ran. `nav-` seems to make sense if end up using the `id="..."` attribute.

philschatz:
(reposting because the comment was collapsed)

@m1yag1 The PR looks great to me! `id="nav-..."`` makes sense.

Just one more question... Could just the link text be used in the tests (rather than creating an attribute or using `:nth`)? That would:

1. help verify that the internationalization code ran successfully
1. wait until the internationalization code finishes before continuing the test
1. let us know when the experience would change for people using a screenreader

kerwinso:
On the front end, this translates to the following links in the top nav bar:
- Search
- About Us
- Give

Issue #1601, opened 2018-03-06
Too many different links, to the same page in a book, is sent to Google Analytics


tomekstach:
Google Analytics should receive only canonical links - now, GA receives all versions of link to the single page in the book.

The problem is with too early sending requests to the GA. `canonicalizePath` in the *src/scripts/modules/media/body/body.coffee* do a replace previous URL with the canonical path in the end of everything - because earlier it doesn't know that `canonicalPath` is not in `currentRoute`.

**The main question is:** how to send requests to the GA after the real canonization of the path? Or maybe will be better to send requests to GA in the end of everything? If yes, how to do that?

tomekstach:
https://github.com/tomekstach/webview/commit/aa663f6067894056f498fd85da224ce00070fbde

I've did only two changes:

1. I added request to the `trackAnalytics` in the end of `canonicalizePath` in the *src/scripts/modules/media/body/body.coffee*
2. I removed `@listenTo(@model, 'change:googleAnalytics', @trackAnalytics)` from the *src/scripts/modules/media/media.coffee*

And it works: request to the GA is sent after the replace URL to the canonical path, but I removed `listenTo` part and it doesn't work properly.

I want to move a part of code, which makes changes in the Google Analytics, to the end of process and then restore 'listenTo' part - but I am not sure how to do this.

tomekstach:
@philschatz Wojtek has sent me a two more links:

- [documentation](https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications)
- [pageVisibilityTracker](https://github.com/krisu7/autotrack/blob/master/docs/plugins/page-visibility-tracker.md)

philschatz:
Oh nice! Both options (using the [autotrack](https://github.com/googleanalytics/autotrack) project and the `urlChangeTracker`) seem promising. I can try creating a Pull Request (or 2) so folks can evaluate and try it out.

philschatz:
Ah, I misunderstood. I think that the `urlChangeTracker` plugin in `autotrack` will still send multiple pageview events since the URL changes multiple times. 

Also, it seems like there is no easy way to add/remove tracking codes when the content changes. Currently, we explicitly call the `pageView` event for all the tracking codes for a piece of content. With `autotrack` it seems like we would need to remember which tracking codes are registered with `ga()` and make sure to de-register them right before `autotrack` sends an event to Google Analytics.

Does that sound right?

I am continuing to look at this double-sending problem (& the commit you mentioned)

[...]

PR #1630, opened 2018-05-09
remove duplicate pageview counts

philschatz:
... by not tracking the canonical URL.

This is not an ideal solution because it causes us to track the non-canonical URL. However, there isn't really a good central place that fires off analytics tracking so this seemed like the minimal fix to address Marketing's concern of double-counting pageviews.

Refs #1601

# What it does with these changes

- `UA79034791` is the org-wide tracking code
- `UA302277982` is the book-specific tracking code

## When going to a page without any version numbers, the following are sent to `ga(...)`:

http://localhost:8000/contents/r-QzKsl_:JEgX-zGG

```
create UA-7903479-1 auto
UA79034791.send pageview /contents/r-QzKsl_:JEgX-zGG
create UA-30227798-2 auto
UA302277982.send pageview /contents/r-QzKsl_:JEgX-zGG
```

## When going to a page directly, the following are sent to `ga(...)`:

http://localhost:8000/contents/r-QzKsl_@7.23:JEgX-zGG@4/Introduction-to-Sociological-R

```
create UA-7903479-1 auto
UA79034791.send pageview /contents/r-QzKsl_@7.23:JEgX-zGG@4/Introduction-to-Sociological-R
create UA-30227798-2 auto
UA302277982.send pageview /contents/r-QzKsl_@7.23:JEgX-zGG@4/Introduction-to-Sociological-R
```

## When clicking `[Next]`, the following are sent to `ga(...)`:

```
create UA-7903479-1 auto
UA79034791.send pageview /contents/r-QzKsl_@7.23:JEgX-zGG/Introduction-to-Sociological-R
create UA-30227798-2 auto
UA302277982.send pageview /contents/r-QzKsl_@7.23:JEgX-zGG/Introduction-to-Sociological-R
```

## When going to just the book uuid URL:

http://localhost:8000/contents/afe4332a-c97f-4fc4-be27-4e4d384a32d8

```
create UA-7903479-1 auto
UA79034791.send pageview /contents/afe4332a-c97f-4fc4-be27-4e4d384a32d8
create UA-30227798-2 auto
UA302277982.send pageview /contents/afe4332a-c97f-4fc4-be27-4e4d384a32d8
```

## When clicking an item in the TableOfContents:

```
create UA-7903479-1 auto
UA79034791.send pageview /contents/r-QzKsl_@7.23:OkLwVQKH/Theoretical-Perspectives
create UA-30227798-2 auto
UA302277982.send pageview /contents/r-QzKsl_@7.23:OkLwVQKH/Theoretical-Perspectives
```

### DIFF
--- a/src/scripts/loader.coffee
+++ b/src/scripts/loader.coffee
@@ -91,8 +91,6 @@ define (require) ->
     if location.search
       router.navigate(Backbone.history.fragment, {replace: true})
 
-    # analytics.sendAnalytics() # Track analytics for the initial page
-
     # Prefix all non-external AJAX requests with the root URI
     $.ajaxPrefilter (options, originalOptions, jqXHR) ->
       if not external.test(options.url)
--- a/src/scripts/loader.coffee
+++ b/src/scripts/loader.coffee
@@ -91,7 +91,7 @@ define (require) ->
     if location.search
       router.navigate(Backbone.history.fragment, {replace: true})
 
-    analytics.sendAnalytics() # Track analytics for the initial page
+    # analytics.sendAnalytics() # Track analytics for the initial page
 
     # Prefix all non-external AJAX requests with the root URI
     $.ajaxPrefilter (options, originalOptions, jqXHR) ->
--- a/src/scripts/helpers/handlers/analytics.coffee
+++ b/src/scripts/helpers/handlers/analytics.coffee
@@ -27,6 +27,10 @@ define (require) ->
 
       require ['analytics'], (ga) =>
 
+        # Uncomment me to log what would have been sent to `ga(...)` (for debugging)
+        # ga = (a,b,c) ->
+        #   console.log('Local Analytics function', a,b,c)
+
         accounts.forEach (account) =>
           if not /^\//.test(fragment) then fragment = '/' + fragment
 
--- a/src/scripts/modules/media/body/body.coffee
+++ b/src/scripts/modules/media/body/body.coffee
@@ -12,6 +12,7 @@ define (require) ->
   SimModal = require('cs!./embeddables/modals/sims/sims')
   template = require('hbs!./body-template')
   settings = require('settings')
+  analytics = require('cs!helpers/handlers/analytics')
   require('less!./body')
 
   embeddableTemplates =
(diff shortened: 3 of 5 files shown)


reedstrm:
Hmm, seems to me we're sending "too early". Can we delay sending until we've rewritten the url to the final with versions form?

tomekstach:
@philschatz please correct me, we will still have more then one address for the one page in the GA statistics? But we should not have many pageViews in the statistics which were caused by redirections?

If it's true, this solution might be ok for us. Of course, it would be ideal to have only one canonical address (in the GA statistics) for the one page.

philschatz:
Analytics now sends the canonical URL. Analytics used to be sent whenever the router was triggered, now it is done explicitly after each page loads.

# Test Plan

This is for @kerwinso . Should it be on the Trello card rather than here?

## Ways to test

There are a couple of ways this could be tested:

1. uncomment the code in `analytics.coffee` which will print out what would have been sent to google analytics (but doesn't actually send)
1. always print what would be sent to google analytics **and** send it to google
1. click around and watch the **Realtime** view in Google Analytics for both cnx.org **and** the book you are viewing
1. Install a chrome extension like https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna that prints out what is being sent to google analytics

## Things to test

- every page type reports to the main cnx.org Google Analytics
  - `/`
  - `/search` add terms, click on links in the left nav bar, click the `[x]` to remove filters, click on a piece of content
  - `/contents`
  - `/browse` click on various subjects
  - `/tos`
  - `/license`
  - `/donate` and click the $10 Donate button
  - `/about` and click the Contact link
  - a 404 page
- Content
  - View a book that we own and verify both codes are sent the pageview
  - View a book that we do not own and verify both codes are sent the pageview
  - View a Page that is not in a book and verify the cnx code is sent
  - View a Page that is not in a book but has an analytics code associated with it and verify the code was sent
  - verify the canonical URL (book shortId + version and Page shortID + version) is the only one sent when visiting
    - a URL containing the full UUID of a book (no version)
    - a URL containing the shortId of a book (no version)
    - a URL containing the full UUID of a Page (no version)
    - a URL containing the shortId of a Page (no version)

tomekstach:
@philschatz I will also try to test it tomorrow on a katalyst01. I am not sure if I will be able to test every single case - but I will try do test as much as I can.

src/scripts/loader.coffee
```diff
@@ -91,7 +91,7 @@ define (require) ->
     if location.search
       router.navigate(Backbone.history.fragment, {replace: true})
 
-    analytics.sendAnalytics() # Track analytics for the initial page
+    # analytics.sendAnalytics() # Track analytics for the initial page
```
reedstrm: commented out? Hmmm.

PR #1631, opened 2018-05-10
Do not strip non-ascii chars from url

mwharrison:
Test plan
=====
Story
-----
AAT, I'd like a webview test that asserts Unicode in URL is allowed

Steps
---
1. Visit:
https://qa.cnx.org/contents/TqqPA4io@1.1:tjWnLaNJ@2 
it redirects to:
https://qa.cnx.org/contents/TqqPA4io@1.1:tjWnLaNJ@2/123-Naprężenie-odkształcenie-i-moduł-sprężystości 
instead of:
https://qa.cnx.org/contents/TqqPA4io@1.1:tjWnLaNJ@2/123-Naprenie-odksztacenie-i-mo
2. Ie.: if the Unicode characters in the URL become stripped out (from the page title), this is a bug.

Expected result
----
- A URL with Unicode characters shows up complete in the address bar.
- Whitespace is replaced with dashes.
- Punctuation is stripped out (ie. question marks, exclamation marks)

### DIFF
--- a/src/scripts/helpers/handlebars/trim.coffee
+++ b/src/scripts/helpers/handlebars/trim.coffee
@@ -5,7 +5,12 @@ define (require) ->
     temp = document.createElement("div")
     temp.innerHTML = str
     str = temp.textContent
-    str.replace(/[\u2000-\u206F\u2E00-\u2E7F\\'!"#$%&()*+,\-.\/:;<=>?@\[\]^_`{|}~\s+]/g, '-')
+    # replace symbols with whitespace
+    str = str.replace(/[\u2000-\u206F\u2E00-\u2E7F\\'!"#$%&()*+,\-.\/:;<=>?@\[\]^_`{|}~\s+]/g, ' ')
+    # remove leading and trailing whitespace
+    str = str.replace(/^\s+|\s+$/g, '')
+    # replace spaces with dashes
+    str = str.replace(/\s+/g, '-')
 
   Handlebars.registerHelper 'trim', (str) -> trim(str)
 
--- a/src/scripts/helpers/handlebars/trim.coffee
+++ b/src/scripts/helpers/handlebars/trim.coffee
@@ -5,7 +5,7 @@ define (require) ->
     temp = document.createElement("div")
     temp.innerHTML = str
     str = temp.textContent
-    str.replace(/\s/g, '-').replace(/[^a-zA-Z0-9 -]/g, '').substring(0, 30)
+    str.replace(/[^\w\s)]/g, '').replace(/\s/g, '-')
 
   Handlebars.registerHelper 'trim', (str) -> trim(str)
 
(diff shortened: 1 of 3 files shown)


mwharrison:
[CHANGE REQUEST: allow non-ascii in URLs](https://trello.com/c/1xB3R6xO/483-change-request-allow-non-ascii-in-urls)

mwharrison:
@reedstrm - it's okay, both this and archive are really still a WIP (which it seems was added automatically). Thanks for pointing out the extra paren, I've fixed that and added a test. Still working on the archive part!

mwharrison:
@reedstrm - I have to use a more comprehensive regex, apparently `/W` matches on non-ASCII characters. 🙄 Thoughts? If you're good with it, let me rebase before you merge so we don't have all these commits going into the repo.

Dantemss:
@BryanEliDimas Michael's regex was stripping out (a giant list of characters) completely and replacing spaces with `-`. The new regex is replacing ALL characters in the (giant list of characters) with `-`. Was this behavior change desired?

Dantemss:
Also maybe add a comment saying what the unicode ranges `\u2000-\u206F` and `\u2E00-\u2E7F` represent (seems to be punctuation characters).

Dantemss:
Also this PR seems to leave unicode characters unescaped in the url (making it an iri I suppose), while https://github.com/Connexions/cnx-archive/pull/556 seems to escape them. Any reason for the different behavior?

If we do decide to escape them all, I believe there are built-in JS functions that will do this, instead of using regexes.

brittweinstein:
Another PR for review: https://github.com/Connexions/cnx-archive/pull/556

kerwinso:
@bryanelidimas please add a test plan to this issue, thanks.

src/scripts/helpers/handlebars/trim.coffee
```diff
@@ -5,7 +5,12 @@ define (require) ->
     temp = document.createElement("div")
     temp.innerHTML = str
     str = temp.textContent
-    str.replace(/\s/g, '-').replace(/[^a-zA-Z0-9 -]/g, '').substring(0, 30)
+    # replace symbols with whitespace
+    str = str.replace(/[\u2000-\u206F\u2E00-\u2E7F\\'!"#$%&()*+,\-.\/:;<=>?@\[\]^_`{|}~\s+]/g, ' ')
```
Dantemss: The `\s+` in this line is probably unnecessary (since the last line condenses all spaces into dashes) but it's a minor thing so I'll approve anyway.

BryanEliDimas:
This PR may or may not still needs @reedstrm's approval.

BryanEliDimas:
@kerwinso Added the test plan to this PR's description.

BryanEliDimas:
This is (more than) ready to be merged :)

PR #1635, opened 2018-05-30
Upgrade MathJax to 2.7.4 and enable SRE for accessibility

pumazi:
Upgrade MathJax to 2.7.4 and enable SRE for accessibility. MathJax was upgraded because the MathJax SRE extension doesn't exist for MathJax 2.6.1.

The [AssistiveMML](https://docs.mathjax.org/en/latest/options/extensions/assistive-mml.html) just exposes MathML to browsers that support reading it... it does not add human-readable alt-text... the [mathjax-sre.js](https://github.com/mathjax/MathJax-a11y/blob/master/docs/README.md#introduction) does that (adds human-readable text).

In order to include `mathjax-sre.js` we had to upgrade MathJax to 2.7

### DIFF
--- a/src/scripts/config.js
+++ b/src/scripts/config.js
@@ -21,7 +21,7 @@
       'backbone-associations': '../../bower_components/backbone-associations/backbone-associations',
 
       // ## MathJax
-      mathjax: '//cdnjs.cloudflare.com/ajax/libs/mathjax/2.6.1/MathJax.js?config=MML_HTMLorMML',
+      mathjax: '//cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.4/MathJax.js?config=MML_HTMLorMML',
 
       // Use Minified Aloha because loading files in a different requirejs context is a royal pain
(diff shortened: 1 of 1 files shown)


philschatz:
@pumazi Unfortunately, I'm having trouble finding the `aria-label` attribute on Math elements... maybe we _do_ need the link for how to resolve `[a11y]/mathjax-sre.js` after all (& maybe explicitly disable `AssistiveMathML`

reedstrm:
ok. Though, the actual _contents_ of the `aria-label` seem … obscure? "collapsed separator separated list of length 1 period" ???

philschatz:
@reedstrm yeah. "Simple" math does seem fine though:

image


The abbreviated alt  text is probably because of the [collapsible a11y plugin](https://docs.mathjax.org/en/latest/options/extensions/a11y-extensions.html#auto-collapse-js) which is enabled by default. We _could_ disable it but I think it's probably enabled so people can skip over long math and explore more if they want.

reedstrm:
Ah, so may actually be usable by the intended audience, then.

PR #1663, opened 2018-06-19
Fix equation numbering style

PatBoj:
image
image


PatBoj:
I've changed the code, it's looks better now.

image

image

Adrianltd:
If you want it to look better, you could center the number within the box (eg. set equal paddings top&bottom and left&right).

Issue #1636, opened 2018-05-31
Improper text wrapping around exercise figures


lwtchu:
https://github.com/Connexions/cnx-recipes/issues/388

Question text wraps incorrectly in some exercises that start with images:

---

![screen shot 2018-05-04 at 3 21 02 pm](https://user-images.githubusercontent.com/28710954/39650856-61a2246c-4faf-11e8-9678-827061a2287a.png)
---
![screen shot 2018-05-04 at 3 20 48 pm](https://user-images.githubusercontent.com/28710954/39650857-61afbb18-4faf-11e8-98e4-d542fccf7194.png)

Book: AP Biology

PatBoj:
Fix it
![image](https://user-images.githubusercontent.com/20907906/42227057-e58224c0-7ee0-11e8-9dc0-6144b0788eab.png)

wilkus27:
Verified on devb https://devb.cnx.org/contents/bDIuMp-w@5.1:w3qFD6Qv@5.1/Critical-Thinking-Questions:

![images_text](https://user-images.githubusercontent.com/40228854/42369406-2a7062e2-810a-11e8-96a4-3ac713ad6bb4.JPG)

wilkus27:
But the problem is still occuring with images of bigger size:
![images_text2](https://user-images.githubusercontent.com/40228854/42370275-9dac3dc4-810c-11e8-83a6-70599a067f33.JPG)

PatBoj:
Maybe it could be fixed in recipes? I think we need `if` statement. If the picture would be too big then the content will go to the next line. What do you thnik @helenemccarron?

Adrianltd:
Or you can remove `float: left;` and style the text paragraphs wherever they're inline to be block and display below image.

openstaxalina:
OpenStax will discuss Monday July 9; will share feedback on July 10 at the latest.

openstaxalina:
For now, let's simply make sure that the content is breaking after images.  We will not have text wrapping around images for our minimum viable product.  We can investigate this further in the future.

PatBoj:
![image](https://user-images.githubusercontent.com/20907906/42682535-25fd8ca4-868b-11e8-82f2-9b085668fade.png)

![image](https://user-images.githubusercontent.com/20907906/42682539-288b4d08-868b-11e8-9fa3-b8d6e83a1c65.png)

openstaxalina:
[...]

Issue #1627, opened 2018-04-10
Find way to not include <p> for inline solutions


brittany-johnson:
In reference to https://github.com/Connexions/webview/pull/1625 

Problem: if we set the solutions to display inline, we may encounter a problem if the solution runs over more than 1 line.

helenemccarron:
closed in favor of https://github.com/Connexions/cnx-recipes/issues/454

PR #1665, opened 2018-06-25
Fix wrapping around an image in exercises

PatBoj:
[Issue #1636](https://github.com/Connexions/webview/issues/1636)
image

### DIFF
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -278,7 +278,7 @@ h1.example-title .text {
 }
 .os-eoc [data-type="exercise"],
 .os-eoc .exercise {
-  p {display: inline; }
+  p { display: inline; }
   [data-type="problem"] > .number {
     .exercise-number();
   }
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -278,10 +278,20 @@ h1.example-title .text {
 }
 .os-eoc [data-type="exercise"],
 .os-eoc .exercise {
-  p { display: inline; }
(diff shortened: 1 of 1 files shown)


helenemccarron:
That is a good point, we do have #1627 that refers to something similar. @PatBoj, you might find some guidance in what Phil and BrittanyJ were thinking. @Stackblocks and @wilkus27 , do you think it is acceptable for this fix to break the number on a different line while we work on the recipes (this will take a while to figure out) or should we try a temporary solution while fixing the markup?

Stackblocks:
@helenemccarron I think that breaking it temporarily while we develop a more robust fix is fine for this one. It doesn't seem to cause any actual harm (read: doesn't change the numbers mysteriously), it just looks a little funky.

helenemccarron:
@PatBoj, let's move on on this PR and work together on fixing the paragraphs later in the recipes.

src/scripts/modules/media/body/content-baked.less
```diff
@@ -278,10 +278,20 @@ h1.example-title .text {
...
     .exercise-number();
   }
+
+  img {
+    float: left;
+    margin-right: 10px;
+  }
+
+  ol {
+    overflow: auto;
```
Adrianltd: Overflow `auto` adds scrollbars when content is overflowing. I think that setting `hidden` there might be better, because we don't want to scroll through elements if this situation arises.

helenemccarron: @Adrianltd good catch! Can you change your review to change requested instead of approved?

Adrianltd: @helenemccarron will do

Issue #1655, opened 2018-06-01
Spacing, text alignment is inconsistent in tables


Stackblocks:
@hwangange commented on [Mon May 21 2018](https://github.com/Connexions/cnx-recipes/issues/399)

There are multiple instances (beyond the ones listed below) of the inconsistent use of "equations" and plain text in the same column, leading to lack of alignment.

Example of spacing being off: 
<img width="630" alt="screen shot 2018-05-21 at 3 20 11 pm" src="https://user-images.githubusercontent.com/12180288/40328026-877a2e2e-5d0a-11e8-8a99-daf52cbc019e.png">

1.2: Physical Quantities and Units

The additional "(=1)" appears on a new line; additionally, it isn't horizontally centered. At the very least, maybe write it as an equation instead of just plain text?

---------------------------------------

Example of text alignment being off:

<img width="924" alt="screen shot 2018-05-21 at 4 12 10 pm" src="https://user-images.githubusercontent.com/12180288/40330169-c1ca2afa-5d11-11e8-8590-4e27ce64a6e1.png">

5.1: Friction

Might look better if the labels ("Static Friction", "Kinetic Friction"), mu symbols, AND decimals are horizontally centered. 

There are plenty of other instances where cells containing column labels align the text to the left but the rest of the cells in the table are horizontally centered ----> inconsistent alignment of text.

Here's another instance. 

<img width="867" alt="screen shot 2018-05-21 at 4 21 12 pm" src="https://user-images.githubusercontent.com/12180288/40330588-06c866f2-5d13-11e8-98d8-48ec46ef30dd.png">

6.1 Rotation Angle and Angular Velocity

"Degree Measure" and "Radian Measure" are aligned left, but the rest of the cells are horizontally centered.

Adrianltd:
@marcjeannott It looks like the alignment issue is caused by the combination of math and plain text. Is that possible to move the (=1) to the math?

Adrianltd:
Alignment in 5.1 is already fixed in baked book
![image](https://user-images.githubusercontent.com/40228907/41624345-b4232544-7415-11e8-9e92-a11c682dcd09.png)

Adrianltd:
And the same in 6.1 
![image](https://user-images.githubusercontent.com/40228907/41624458-0c4dcb98-7416-11e8-968e-daaf2c9a4347.png)

PR #1670, opened 2018-06-26
Issue 1655 spacing textalign in tables

Adrianltd:
I had to override MathJax, I'm still not really sure if that's the correct way. I've checked few tables in other books, seems that it didn't affect them badly, but surely there's a need for QA in here.
#1655 issue link

### DIFF
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -224,7 +224,12 @@ h1.example-title .text {
 .os-table {
   margin: @line-height-computed 0;
   table {
-     margin: 0;
+    margin: 0;
+    tr {
+      td {
+        vertical-align: middle;
+      }
+    }
   }
   .os-caption-container {
     padding: 8px;
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -220,6 +220,16 @@ h1.example-title .text {
   }
 }
 
+
+//
+// MathJax_Display override
+// --------------------------------------------------
+.MathJax_Display {
+  width: auto !important;
+  float: left !important;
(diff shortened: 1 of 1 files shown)


Adrianltd:
image
image
image

helenemccarron:
@philschatz I think this is an acceptable change for now but I'm not sure we are approaching MathJax the proper way. Is there something else we should consider before hacking at the MathJax CSS?

src/scripts/modules/media/body/content-baked.less
```diff
@@ -220,12 +220,28 @@ h1.example-title .text {
   }
 }
 
+
+//
+// MathJax_Display override
+// --------------------------------------------------
+.MathJax_Display {
+  width: auto !important;
```
philschatz: Can you include a description of why the override occurs? Also, so that it shows up in dev tools, I think the following format works: `width: auto !important /*because ....*/;`

helenemccarron:
I can do some research on where the CSS for MathJax is being kept, I was hoping you knew, @philschatz :)

src/scripts/modules/media/body/content-baked.less
Adrianltd: I'm not sure if I understand you correct. That's the only possible way I've found to override default MathJax styles.

I also thought of putting all mathjax elements into some containers and style the container + reset all mathjax displaying properties (not the math elements structure itself) but someone said that's not the best idea.

I think we should be able to style MathJax elements through the MathJax config file, am I right?

Which of those might be the best choice?

Adrianltd:
@helenemccarron this can be merged, right? and we'll get back to the mathjax styling topic on thursday

helenemccarron:
@Adrianltd yes, this is ready

Issue #1669, opened 2018-06-26
Polish text on English webview


wilkus27:
When language of browser is non-English, part of text on the page is displayed in Polish (I don't check browsers with other languages):

![pol](https://user-images.githubusercontent.com/40228854/41901621-adb4a874-7931-11e8-8152-8048738b2cbb.JPG)

What is more, the same text on different pages, one time is in English and other time in Polish:

![links2](https://user-images.githubusercontent.com/40228854/41901847-457bb634-7932-11e8-9c02-7bf1bcc81abb.JPG)
![links3](https://user-images.githubusercontent.com/40228854/41901857-4a7f22ec-7932-11e8-9304-3cfb5746a3e5.JPG)

Other example: 

![links](https://user-images.githubusercontent.com/40228854/41901954-8c5ed644-7932-11e8-89d8-1cbb4c783d6f.JPG)

Licensing is all in Polish, on Term of Use page there are a few words in Polish and Accessibility Statement is all in English.

helenemccarron:
@PiotrKozlowski feel free to create a new pull request with your change!

PiotrKozlowski:
For example we can see footer in polish on home or about page but when we open any book then footer will be in english.

PiotrKozlowski:
@helenemccarron Should I do something here?

helenemccarron:
ooh, that's interesting. @reedstrm or @philschatz do you have any insight on this?

philschatz:
@PiotrKozlowski & @helenemccarron : oh, I've seen this error too (I thought it was just one of my browser extensions but I guess not). Ah, l20n is running `JSON.parse(element.getAttribute('data-l10n-args') || null)];` which causes the exception.

And the element looks like the following:

`<span class="title" data-l10n-id="textbook-view-page-in-books" data-l10n-args="{&quote;numbooks&quote;: }">`

It seems `numbooks:` never has a number set to it which is causing the "Invalid JSON".

# Screenshot

![image](https://user-images.githubusercontent.com/253202/42327441-af28e85a-8039-11e8-8b5c-73a8445f1e24.png)

PiotrKozlowski:
It's related to 
`webview\src\scripts\modules\media\tabs\contents\contents-template.html`
or
`dictionary.ftl` files

I need to check how l20n works to solve this

openstaxalina:
I think this is resolved by the attached PR.  If so, this can likely be assigned to Natalia for testing.

[...]

PR #1687, opened 2018-07-06
issue-1669-polish-text-on-english

PiotrKozlowski:
#1669 

Revert "Wait to load book count (#1621)"
This reverts commit dff42c00ca49ceb46272c4cac2410ce49e227fa2.

I believe that in previous version there was no translations for this lines so that code was working but now this is cousing an error so we reverted this. Thanks to @aiwenar 

cc @sisiz

### DIFF
--- a/src/scripts/modules/media/tabs/contents/contents.coffee
+++ b/src/scripts/modules/media/tabs/contents/contents.coffee
@@ -68,7 +68,11 @@ define (require) ->
       booksIn: () ->
         @model?.get('booksContainingPage')
       numBooks:() ->
-        @model?.get('booksContainingPage')?.length
+        hits = @model?.get('booksContainingPage')?.length
+        if hits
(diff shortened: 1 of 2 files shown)


src/scripts/modules/media/tabs/contents/contents.coffee
```diff
@@ -67,15 +67,8 @@ define (require) ->
...
-        hits = @model?.get('booksContainingPage')?.length
-        return unless hits?
-        if hits is 0
-          "This page is not in any books."
-        else if hits is 1
-          "This page is in this book:"
-        else
-          "This page is in #{hits} books:"
+      numBooks:() ->
+        @model?.get('booksContainingPage')?.length
```
philschatz: Should this return `0` if `@model` or `@model.get('booksContainingPage')` are undefined?

Issue #1686, opened 2018-07-05
College Physics Chapter outlines need to be designed and styled 


helenemccarron:
@openstaxalina, could you provide specs or the design guidance? Currently, they are very confusing.
![2018-07-05_11-06-21](https://user-images.githubusercontent.com/8514591/42335297-86a3c236-8045-11e8-8cc3-de313eedbc70.png)

helenemccarron:
This would be using styles matching webivew or should we try to replicate a similar layout (centered text, block headers etc?)

openstaxalina:
@helenemccarron styles matching webview, but using the screenshot as a guideline for moving the bullets closer to their section header, and making the entire outline more compact.

helenemccarron:
![2018-07-05_13-20-42](https://user-images.githubusercontent.com/8514591/42342967-495279e6-805d-11e8-91ce-0a4ea839f4b8.png)
@openstaxalina approved :D Please make the chapter outline look like the above

PatBoj:
![image](https://user-images.githubusercontent.com/20907906/42682461-f405cea0-868a-11e8-9357-6cf034eabf46.png)

wilkus27:
Veriffied on katalyst01 (https://katalyst01.cnx.org/contents/Ax2o07Ul@11.32:HR_VN3f7@3/Introduction-to-Science-and-th):
![chapter_outline](https://user-images.githubusercontent.com/40228854/43128028-3a24a22c-8f31-11e8-893f-d21e58628299.JPG)

PR #1690, opened 2018-07-06
Fixed design of College Physics Chapter outlines

PatBoj:
It was from #1686.
image

### DIFF
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -33,6 +33,13 @@ iframe {
   text-align: center;
   font-weight: normal;
 }
+
+.os-chapter-outline {
+  ul {
+    margin-top: 0;
+  }
+}
+
 // It is already displayed at the top of the page in blue
 // the `>` is because collated titles (like Key Concepts) contain descendent
 // document-title elements.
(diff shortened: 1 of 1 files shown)


src/scripts/modules/media/body/content-baked.less
```diff
@@ -178,8 +185,6 @@ h1.example-title .text {
 .note     { .make-block(note); }
 [data-type="example"],
 .example  { .make-block(example); }
-[data-type="abstract"],
```
Adrianltd: abstract = chapter outline? does it need to happen globally?
if that code doesn't affect anything else badly, I approve

src/scripts/modules/media/body/content-baked.less
```diff
@@ -178,8 +185,6 @@ h1.example-title .text {
...
-.abstract { .make-block(note); }
```
PiotrKozlowski: This looks like some global mixin for `.abstract`. Should we remove it or just reset in `.os-chapter-outline`? @helenemccarron

helenemccarron:
The learning objectives and chapter outlines are confusing. Some of the chapter outline (using the class/data-type `abstract`) are generated in the content, some (using the class 'chapter-outline` or `learning-objectives`) are created directly in the content. Visually, there shouldn't be a difference, we tried to unify those in the recipes but there is such a descrepancy between books, it's really hard to catch all of them. I would recommend testing several books (biology, physics, economics,  statistics) this should give us a good idea of the range of markup that needs to be supported. Hopefully, this should be as easy as looking at the first chapter of all those books and confirm the fix proposed by Patryk is appropriate.

PatBoj:
@helenemccarron everything I've checked physics, economics, statistics and biology books and they look fine to me.

Issue #1682, opened 2018-07-03
Two types of items


wilkus27:
In Unit 5-29.3 https://devb.cnx.org/contents/GFy_h8cu@10.120:TIORyfla@5/293-Amphibians in Link to Learning box there are two types of items next to each other. One should be removed.
![items](https://user-images.githubusercontent.com/40228854/42226823-378473dc-7ee0-11e8-83dd-404e3d54ada0.JPG)

Adrianltd:
This list is created with unordered list `ul` and then there's numbering inside of the `li` elements (e.g.`1: The evolution from fish to earliest tetrapod`).

I suggest that we keep the numbers instead of the dots, because the title says that we have exactly 5 videos from Animal Planet, and that's something that will not change.
We should transform that into ordered list during baking process and remove numbering from the content. 

What do you think @helenemccarron, @openstaxalina?

helenemccarron:
Yes, I agree, disable the ul styling.

openstaxalina:
Yep, sounds good @Adrianltd.  cc @helenemccarron

helenemccarron:
@openstaxalina upon further review, I think we should change the ul to an ol rather than hard code the numbers. If we remove styling for unstyled lists, we might cause issues for all other lists. 

![2018-07-06_08-43-25](https://user-images.githubusercontent.com/8514591/42382082-d1014f98-80f8-11e8-86bd-61c4ff1dd5bd.png)

openstaxalina:
Actually, this is no longer an issue on production:
![screen shot 2018-07-06 at 9 16 06 am](https://user-images.githubusercontent.com/7863733/42383588-41593d4c-80fd-11e8-9b40-6b807415285a.png)

There is another issue where the videos are spilling outside the note container.  I am going to talk with @Stackblocks today about a possible content refresh on DevB so we can ensure we are testing updated content.

PR #1691, opened 2018-07-06
Issue 1682 two types of items

Adrianltd:
#1682 
removed list-style from ul/ol, becase numbering is hardcoded already


helenemccarron:
@PiotrKozlowski Good point. I didn't realize the content was hard coded. This would also create regressions for other lists.

helenemccarron:
#1682 was closed, closing the PR

PR #1714, opened 2018-07-12
Fix improper text wrapping around exercise figures

PatBoj:
#1636 I've breaking the line after every image that appears in the exercise.
image

image

### DIFF
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -302,14 +302,9 @@ h1.example-title .text {
   }
 
   img {
-    float: left;
-    margin-right: 10px;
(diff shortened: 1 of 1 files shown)


src/scripts/modules/media/body/content-baked.less
```diff
@@ -302,14 +302,9 @@ h1.example-title .text {
...
 
   img {
-    float: left;
-    margin-right: 10px;
-  }
-
-  ol {
-    overflow: hidden;
+    display: block;
+    padding-bottom: 1em;
```
Adrianltd: Use margin instead of padding, so if anyone adds background/border or anything to this element, it will look ok too.

Issue #1701, opened 2018-07-09
No margin in table content


uskka:
Where: Every book on devb. Screen from devb / College Physics / 2.7-Falling-Objects 
https://devb.cnx.org/contents/Ax2o07Ul@9.98:nkhtuBIO@6/27-Falling-Objects

Description: No margin in all of the tables content. 

![screenshot-devb cnx org-2018 07 09-10-37-51](https://user-images.githubusercontent.com/40228918/42439917-79ef79a0-8364-11e8-9bf8-26d6791e395e.png)

openstaxalina:
We would like to add padding so the content isn't flush with the border lines of the table.

uskka:
Works fine. Tested on katalyst01.
![screenshot-katalyst01 cnx org-2018 07 24-10-53-33](https://user-images.githubusercontent.com/40228918/43127647-3ff38dcc-8f30-11e8-9b6a-d5b444d76cec.png)

PR #1715, opened 2018-07-12
add padding left to <td> (same as for <th>)

PiotrKozlowski:
issue #1701 

image

### DIFF
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -240,7 +240,7 @@ h1.example-title .text {
     tr {
       td {
         vertical-align: middle;
-        padding: 1em 0em;
+        padding: 1em 0 1em 8px;
       }
     }
   }


src/scripts/modules/media/body/content-baked.less
```diff
@@ -240,7 +240,7 @@ h1.example-title .text {
     tr {
       td {
         vertical-align: middle;
-        padding: 1em 0em;
+        padding: 1em 0 1em 8px;
```
helenemccarron: should we use only one unit? Maybe just `em`?

PiotrKozlowski: So `<th>` is set with `padding: @table-cell-padding;` which stands for `padding: 8px;`
Propably the best way will be to set `td` with the same way?

helenemccarron: Is it possible to change `padding: 1em 0 1em 8px;` to `padding: @table-cell-padding 0 @table-cell-padding;`? Or maybe we simply use ` padding: 1em 0 1em 1em;`?

Issue #1704, opened 2018-07-09
Intro to Business: Terms and Index Terms are Bolded


Bkblodget:
both <term> and <glossary> items appear as bolded text in webview. Only <glossary> terms should appear in bolded text, as in the PDF.

I believe this happens because webview lacks an Index.

openstaxalina:
This is most likely happening for all titles.

openstaxalina:
@Bkblodget will provide any known examples for testing purposes.  These changes will apply to all books.

brittany-johnson:
University Physics V 3 Key Terms section on production
![screen shot 2018-07-10 at 10 14 01 am](https://user-images.githubusercontent.com/24909007/42519696-39e22c2e-842a-11e8-8239-5c317a4f4e8f.png)

Intro to Business Key Terms Section on devb
![screen shot 2018-07-10 at 10 14 10 am](https://user-images.githubusercontent.com/24909007/42519697-39f58d3c-842a-11e8-8472-1bfebd21019e.png)

Index U Physics 
<img width="1373" alt="screen shot 2018-07-10 at 10 15 34 am" src="https://user-images.githubusercontent.com/24909007/42519704-3d3801f0-842a-11e8-8964-92a4dd037e0e.png">

Index Intro business 
<img width="1345" alt="screen shot 2018-07-10 at 10 15 24 am" src="https://user-images.githubusercontent.com/24909007/42519705-3d51f0f6-842a-11e8-90bc-a4bdf5f903e7.png">


The key terms in intro business appear to be displaying correctly, and it has an index. I used UPhysics as another example that's on production since the syles are universal. Since this issue was created a day ago, something weird may be going on if youre not able to see this. 

 [Intro business on dev b](https://devb.cnx.org/contents/Mmq0W-88@12.1:SBfkOk9C@12.1/Key-Terms)

Let me know if you're still not seeing what is in the screenshots! @Bkblodget cc: @Stackblocks

Bkblodget:
@brittany-johnson sorry for the lack of clarification - this issue actually refers to terms that are in the body portion of the content. For instance, in the screenshot below **Walmart** is a term tag and should not appear bolded. 

![screen shot 2018-07-10 at 10 28 23 am](https://user-images.githubusercontent.com/11299035/42520560-2a02ae08-842c-11e8-8be4-b0295c9505ea.png)

brittany-johnson:
@Bkblodget AHHH! Makes sense now, thank you!

[...]

PR #1720, opened 2018-07-17
Issue-1704

brittany-johnson:
https://github.com/Connexions/webview/issues/1704

### DIFF
--- a/src/scripts/modules/media/body/content-common.less
+++ b/src/scripts/modules/media/body/content-common.less
@@ -130,6 +130,10 @@ figure[data-orient="vertical"] {
   font-weight: bold;
 }
 
+.no-emphasis {
+  font-weight: normal;
+} //to ensure that all terms with this tag will display unbolded
+
 // Teacher's Edition
 .os-teacher {
   display: none;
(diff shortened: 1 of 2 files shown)


src/scripts/modules/media/body/content-common.less
```diff
@@ -130,6 +130,10 @@ figure[data-orient="vertical"] {
   font-weight: bold;
 }
 
+.no-emphasis {
```
helenemccarron: is this intended to be a change that applies to all of webview? Or should it be moved to content-baked.less?

brittany-johnson: I put it in `content-common.less` because I figured that anytime there is a `.no-emphasis` it shouldn't be bolded, BUT I keep forgetting about the testing implications. I should probally apply this to only baked books correct, no matter how small the change? @helenemccarron 
cc @Stackblocks

helenemccarron: I seem to remember going back and forth on this with Alina, I may be wrong though. cc @openstaxalina

PR #1727, opened 2018-07-18
issue-1712

brittany-johnson:
style top titled table titles


src/scripts/modules/media/body/content-common.less
```diff
@@ -426,6 +426,14 @@ table {
   }
 }
 
+.os-table {
+  .os-table-title {
+    text-align: center;
+    font-weight: bold;
+    .top-titled, {}
```
tomjw64: Is this an unused empty block?

Issue #1728, opened 2018-07-19
left pane shows loading message 'This page is not in any books' when page is in a book


PiotrKozlowski:
Full desc:

https://trello.com/c/WQ4KGS5N/567-bug-left-pane-shows-loading-message-this-page-is-not-in-any-books-when-page-is-in-a-book

@kerwinso @philschatz

kerwinso:
Moved this to `nice.naga` milestone so we can track for 7/30 release.

Stackblocks:
Reported as verified by @kerwinso, closing.

PR #1729, opened 2018-07-19
Fix Issue 1728 - left pane shows loading message 'This page is not in any books' when page is in a book

PiotrKozlowski:
#1728 

Add condition to show  div with `This page is not in any books` only when book is fully loaded


kerwinso:
Kerwin to test this PR on qa -- if it passes, this PR should be merged to `master`, not to `katAW.nice.naga`

PR #1734, opened 2018-07-23
KatAW.ornery.orthrus

helenemccarron:
Dates:
    Code freeze (no new features): 2018-07-19
    Code deployed to staging server: 2018-07-20
    Testing on staging server: 2018-07-23 => 2018-07-27
    RC Review meeting: 2018-07-24
    Code Released to Production: 2018-07-30

### DIFF
--- a/src/scripts/modules/media/body/content-common.less
+++ b/src/scripts/modules/media/body/content-common.less
@@ -1,3 +1,4 @@
+@table-cell-padding: 1em;
 // Styles for the contents of a page.
 // This should **only** be imported inside a containing selector like:
 //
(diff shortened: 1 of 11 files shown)


src/scripts/modules/media/body/content-common.less
```diff
@@ -1,3 +1,4 @@
+@table-cell-padding: 1em;
 // Styles for the contents of a page.
```
helenemccarron: @PiotrKozlowski can we move this variable to content-baked.less, I am worried about regressions if we keep it in common.

src/scripts/modules/media/body/content-common.less
```diff
@@ -345,6 +346,16 @@ table {
...
+  ul {
+    padding: 0;
+    margin: 0;
+    &[data-bullet-style="none"] {
+      li {
+        list-style: none;
+      }
+    }
+  }
+
```
helenemccarron: I am not sure why this is here but can we move it `content-baked.less` as well to avoid any regressions? @PiotrKozlowski @Adrianltd @PatBoj

Issue #1739, opened 2018-07-24
Issue with big tables


Stackblocks:
@Stackblocks commented on [Fri Jul 20 2018](https://github.com/Connexions/cnx-recipes/issues/488)

@uskka commented on [Mon Jul 09 2018](https://github.com/Connexions/webview/issues/1705)

Description: When Contents are shown and table is too big, part of the table content is cut out of the screen. There should be mechanism that prevents such behaviour.

Where: Screen from devb/ Accounting / B | Present Values Table Set Rotated (Landscaped at 90°) as Art and Table
https://devb.cnx.org/contents/l6eNxxA-@2.6:YykWtsbi@3/B--Present-Values-Table-Set-Ro

![screenshot-devb cnx org-2018 07 09-15-58-18](https://user-images.githubusercontent.com/40228918/42454985-469448e2-8391-11e8-8ea6-9be424b92938.png)
![screenshot-devb cnx org-2018 07 09-15-58-56](https://user-images.githubusercontent.com/40228918/42454988-4836e6d2-8391-11e8-84a9-3f8f2ca03c3d.png)

---

@helenemccarron commented on [Mon Jul 09 2018](https://github.com/Connexions/webview/issues/1705#issuecomment-403528455)

@openstaxalina I am not sure what we can do CSS wise to fix this type of issue. 

---

@openstaxalina commented on [Mon Jul 09 2018](https://github.com/Connexions/webview/issues/1705#issuecomment-403635479)

@helenemccarron can this be fixed with JavaScript?  I would like to try to include this with Assignable Webview.  The screen won't scroll right, so the content is completely inaccessible/illegible for the user.

---

@PiotrKozlowski commented on [Tue Jul 10 2018](https://github.com/Connexions/webview/issues/1705#issuecomment-403723587)

You can try something like this:
```
.os-table{
  overflow: scroll;
  overflow-y: hidden;
}
```
Then in mobile you will be able to scroll table by swiping and on desktop you can scroll horizontally with `shift + scroll`. Eventually we can add js script which will allow to scroll tables horizontally on desktop also by swiping with mouse.

---

@openstaxalina commented on [Tue Jul 10 2018](https://github.com/Connexions/webview/issues/1705#issuecomment-403898247)

[...]

PR #1741, opened 2018-07-24
enable swiping big tables with mouse for desktop users

PiotrKozlowski:
#1739 

How it works:
If user is not on mobile device then script will look for tables which width is greater than their parents widths. In that case script will listen to `mousedown`, `mousemove`, etc. events adn apply proper `target.scrollLeft` value.

@philschatz 
I will ned your help to find the best place to include this script. 

For now it is loading after content is loaded - which is good. But when user will change page by navigating trough menu then script should load one more time. 

Propably it should be import as a function which will be called after page change but I can't find a place for it.

### DIFF
--- a/src/scripts/helpers/swipe-tables.js
+++ b/src/scripts/helpers/swipe-tables.js
@@ -1,0 +1,46 @@
+if( !/Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent) ) {
+  let tablesToSwipe = Array.prototype.filter.call(document.getElementsByClassName('os-table'), function(el){
+    if (el && (el.offsetWidth < el.getElementsByTagName('table')[0].offsetWidth)) {
+      return el
+    }
+  })
+  
+  if (tablesToSwipe.length > 0) {
+    tablesToSwipe.forEach(function(el){
+      el.classList.add('swipe-table')
+      el.addEventListener('mousedown', function(){
+        startSwiping(el)
+      })
+      el.addEventListener('mouseup', function(){
+        stopSwiping(el)
+      })
+      el.addEventListener('mouseleave', function(){
+        stopSwiping(el)
+      })
+    })
+  }
+  
+  let tableLeftPos, startSwipingPos, swipeRange
+  
+  function handleSwipe (target) {
+    let mouseX = window.event.clientX
+    let swipeDistance = startSwipingPos - mouseX
+    if (swipeDistance <= swipeRange && swipeDistance >= (-1 * swipeRange)) {
+      target.scrollLeft += swipeDistance
+    }
+  }
+  
+  function startSwiping (target) {
+    offsets = target.getBoundingClientRect()
+    tableLeftPos = offsets.left
+    startSwipingPos = window.event.clientX
+    swipeRange = target.getElementsByTagName('table')[0].offsetWidth - target.offsetWidth
+    target.onmousemove = function(e) {
+      handleSwipe(target)
+    }
+  }
+  
+  function stopSwiping (target) {
+    target.onmousemove = function(){}
+  }
+}
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -240,7 +240,17 @@ h1.example-title .text {
 
 // Tables
 // --------------------------------------------------
+.swipe-table {
+  -webkit-touch-callout: none; /* iOS Safari */
+    -webkit-user-select: none; /* Safari */
+     -khtml-user-select: none; /* Konqueror HTML */
+       -moz-user-select: none; /* Firefox */
+        -ms-user-select: none; /* Internet Explorer/Edge */
+            user-select: none; /* Chrome / Opera */
+  cursor: -webkit-grab;
+}
 .os-table {
+  overflow-x: auto;
   margin: @line-height-computed 0;
   .os-table-title {
     text-align: center;
--- a/src/scripts/modules/media/body/body.coffee
+++ b/src/scripts/modules/media/body/body.coffee
@@ -199,6 +199,8 @@ define (require) ->
             $el = $(el)
             $el.css('counter-reset', 'list-item ' + $el.attr('start'))
 
+          require(['helpers/swipe-tables'])
+
           # # uncomment to embed fake exercises and see embeddable exercises in action
           # @fakeExercises($temp)
 


src/scripts/helpers/swipe-tables.js
```diff
@@ -0,0 +1,50 @@
+(function () {
+  'use strict';
+
+  var tableLeftPos, startSwipingPos, swipeRange;
+    
+  function handleSwipe (target) {
+    var mouseX = window.event.clientX;
```
philschatz: It seems like `target.event` might also provide this object. What is the advantage of using the global `window.event`?

See main review comment for alternate suggestions that may not need this code

src/scripts/helpers/swipe-tables.js
```diff
@@ -0,0 +1,50 @@
...
+      target.scrollLeft += swipeDistance;
+    }
+  }
+  
+  function startSwiping (target) {
+    var offsets = target.getBoundingClientRect();
+    tableLeftPos = offsets.left;
+    startSwipingPos = window.event.clientX;
+    swipeRange = target.getElementsByTagName('table')[0].offsetWidth - target.offsetWidth;
+    target.onmousemove = function() {
```
philschatz: I think that using `target.onmouse...` makes it difficult for other parts of the code to attach events. Could [addEventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) be used instead?

See main review comment for alternate suggestions that may not need this code

src/scripts/helpers/swipe-tables.js
```diff
@@ -0,0 +1,50 @@
...
+    var tablesToSwipe = Array.prototype.filter.call(document.getElementsByClassName('os-table'), function(el){
+      if (el && (el.offsetWidth < el.getElementsByTagName('table')[0].offsetWidth)) {
+        return el;
+      }
+    });
+    
+    if (tablesToSwipe.length > 0) {
+      tablesToSwipe.forEach(function(el){
+        el.classList.add('swipe-table');
+        el.addEventListener('mousedown', function(){
```
philschatz: Will these event listeners need to be removed when the page changes or if the user resizes the screen?

See main review comment for alternate suggestions that may not need this code

src/scripts/helpers/swipe-tables.js
```diff
@@ -0,0 +1,50 @@
...
+      handleSwipe(target);
+    };
+  }
+  
+  function stopSwiping (target) {
+    target.onmousemove = null;
+  }
+
+  if( !/Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent) ) {
+    var tablesToSwipe = Array.prototype.filter.call(document.getElementsByClassName('os-table'), function(el){
```
philschatz: Could this be written more succinctly using `[...document.querySelectorAll('.os-table')].filter(el => ...)` syntax? (or coffeescript, but might not be worth learning)

See main review comment for alternate suggestions that may not need this code

PiotrKozlowski:
@philschatz This is implemented in css with `overflow-x: auto` for `.os-table`. 
`swipe-tables.js` is enabling users to swipe tables with their mouse (on mobile its standard) instead of using scrollbar at the bottom (it's working like on the trello boards where you can click and swipe whole board with mouse). This behaviour is also implmented in Apple magic mouse as a standard.

I will implement your suggestions.

src/scripts/helpers/swipe-tables.js
PiotrKozlowski: `Uncaught TypeError: document.querySelectorAll(...).filter is not a function`
```
Array.isArray(document.querySelectorAll('.os-table'))
false
```
I'ts not an Array so I'am not sure if there is another work around for this than calling `Array.prototype....call`

src/scripts/helpers/swipe-tables.js
PiotrKozlowski: Not working for me `Uncaught TypeError: Cannot read property 'clientX' of undefined`

philschatz: ah, it's because the function on line 14 does not take an argument. I think something like the following would work:

```
target.onmousemove = function(t) {
  handleSwipe(t)
}

handleSwipe(t) {
  console.log(t.clientX)
}
```

I had assumed that `target` was the 1st argument to the function defined at `target.onmousemove = function ...`

src/scripts/helpers/swipe-tables.js
philschatz: `[...xyz]` converts an Iterable (`xyz`) into an array using the [Array Spread](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) syntax. I think `[...document.querySelectorAll('.os-table')].filter(el => ...)` should work

src/scripts/modules/media/body/content-baked.less
```diff
@@ -240,7 +240,17 @@ h1.example-title .text {
 
 // Tables
 // --------------------------------------------------
+.swipe-table {
```
philschatz: Maybe I'm misunderstanding but if the JS _just_ added a `.is-not-mobile` class on the root element then couldn't we just add something like the following:

```css
.is-not-mobile {
  overflow-x: scroll; /* or : auto; , my CSS is a bit rusty */
}
```

Then, there would be no need to make sure JS listeners were added whenever the page content changed, right?

PiotrKozlowski: You have to add listeners to enable users to swipe tabels. It's all about allowing users to swipe instead of using normal scrollbar

src/scripts/helpers/swipe-tables.js
PiotrKozlowski: This would work but travis... :D 
```
29 |    var tablesToSwipe = [...document.getElementsByClassName('os-table')].filter(el => {
                                  ^ [W119] 'spread/rest operator' is only available in ES6 (use 'esversion: 6').
     29 |    var tablesToSwipe = [...document.getElementsByClassName('os-table')].filter(el => {
                                                                                            ^ [W119] 'arrow function syntax (=>)' is only available in ES6 (use 'esversion: 6').
```

src/scripts/modules/media/body/content-baked.less
philschatz: ok, thanks for the explanation. Could we still use the event delegation rather than attaching these listeners to the DOM nodes directly? That would help with the issue described in the Pull Request description (needing to re-add the listener when the page changes)

philschatz:
Rather than creating a new View, could the events be added to https://github.com/Connexions/webview/blob/master/src/scripts/modules/media/body/body.coffee#L41-L46 ? For example, something like:

```coffee
events:
  # Keep Existing event handlers here.
  
  'keydown' : 'keyAction'
  'mousedown .os-table' : 'startSwiping'
  'mouseup .os-table' : 'stopSwiping'
  'mouseleave .os-table' : 'stopSwiping'
  'mousemove .os-table' : 'handleSwipe'
```

The `MediaBodyView` already adds/removes event listeners when the content changes so the swipe events would also be added and removed when the content changes.

To convert the JavaScript functions that have been written, https://js2.coffee might be useful.

PiotrKozlowski:
@philschatz Thanks for your engagement. I moved events and functions, travis also works now correctly.
Is there something else which I can improve?

philschatz:
Awesome! I did not try it myself but it looks like all of the concerns were addressed!

philschatz:
Awesome! I did not try it myself but it looks like all of the concerns were addressed!

Issue #1650, opened 2018-06-01
Inconsistent spacing between number and first word of question, in Problems & Exercises


Stackblocks:
@hwangange commented on [Mon May 21 2018](https://github.com/Connexions/cnx-recipes/issues/403)

There are seemingly sporadic instances where no space exists between a question's number and first word. Like question 20 in Ch. 2 Kinematics, Problems & Exercises. 
<img width="852" alt="screen shot 2018-05-21 at 3 54 28 pm" src="https://user-images.githubusercontent.com/12180288/40329472-4497bb1c-5d0f-11e8-8343-a5157eb41879.png">

And question 26 (also Ch. 2).
<img width="737" alt="screen shot 2018-05-21 at 3 54 58 pm" src="https://user-images.githubusercontent.com/12180288/40329499-56f6bbb4-5d0f-11e8-82c1-c9e98ff81c24.png">

And question 59 (also Ch. 2).
<img width="142" alt="screen shot 2018-05-21 at 3 55 27 pm" src="https://user-images.githubusercontent.com/12180288/40329521-6704d888-5d0f-11e8-915b-58089007dc6f.png">

And questions 13-16, Ch.3 Two-Dimensional Kinematics, Conceptual Questions.

<img width="462" alt="screen shot 2018-05-21 at 4 01 16 pm" src="https://user-images.githubusercontent.com/12180288/40329767-3d836528-5d10-11e8-9c13-e6d96a299a16.png">

also in Ch.3 Test Prep, etc.

openstaxalina:
This is a duplicate of the issue in cnx-recipes referenced above; closing.

openstaxalina:
Actually, we are not sure what repo this should be in.  Helene will investigate.

helenemccarron:
This is not a recipe issue, the baked content does not contain extra spaces.

helenemccarron:
In the JSON found in https://archive-devb.cnx.org/contents/8d04a686-d5e8-4798-a27d-c608e4d0e187@14.60:OF8r0KAI@14.60/Problems--Exercises.json: 
`<a class=\"os-number\" href=\"/contents/5b05c06e-0cbd-5019-be06-2abd4de9909d#fs-id1770908-solution\">61</a>    <p id=\"import-auto-id1761648\">Using approximate values` 

[...]

PR #1749, opened 2018-07-26
issue 1650 imitate space with padding for .os-number inside eoc, eob

PiotrKozlowski:
#1650
This PR propably should be merged after `https://github.com/Connexions/oer.exports/issues/3714` will be resolved?

@cc @openstaxalina


helenemccarron:
@PiotrKozlowski we can proceed with the CSS while the content is being worked on. The result will still be uneven (either one space or two spaces rather than currently no space and one space).

PR #1772, opened 2018-07-31
Issue 1759

damondemar:
Gave button dialog role and added dialog title

### DIFF
--- a/src/scripts/modules/media/header/header-template.html
+++ b/src/scripts/modules/media/header/header-template.html
@@ -35,7 +35,9 @@
       <div class="downloads">
         {{#or hasDownloads buyLink}}
           {{#is mediaType 'application/vnd.org.cnx.collection'}}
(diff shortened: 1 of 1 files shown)


src/scripts/modules/media/header/header-template.html
```diff
@@ -35,7 +35,9 @@ <h2>
       <div class="downloads">
         {{#or hasDownloads buyLink}}
           {{#is mediaType 'application/vnd.org.cnx.collection'}}
-            <button class="btn" tabindex="0" type="submit" data-l10n-id="textbook-view-btn-get-this-book">Get This Book!</button>
+            <button class="btn" tabindex="0" type="submit" data-l10n-id="textbook-view-btn-get-this-book" role="dialog" aria-labelledby="dialog1Title">Get This Book!
```
tomjw64: I don't really know what this does, but is the "Get This Book!" string supposed to be duplicated on both lines?

damondemar: No it's not. I meant to remove the first "Get This Book!" but forgot

PR #1787, opened 2018-08-02
Fixed numbering exercises to be inline

PatBoj:
Before:
image

After:
image


PatBoj:
@helenemccarron Is everything ok with my solution? Should it look like on screen-shots? If yes I could move this issue to testing.

helenemccarron:
@PatBoj yes, that looks good!

PR #1793, opened 2018-08-02
Added files necessary to run webview dev environment in a docker container

m1yag1:
* Added Dockerfile, .dockerignore, docker-compose.yml, and docker-compose.override.yml
* Renamed nginx.development.conf to nginx.dev.conf and moved it into `./conf` directory
* Added supervisord.dev.conf to keep the docker container from closing out immediately
* Added Docker instructions to the README.md
* Updated `Gruntfile.coffee` to include new location of the ngix configuration file.


m1yag1:
Looks like I have a conflict. I'll fix that soon. I also need to add an override for the nginx config file b/c someone may want to tweak that too.

Issue #1703, opened 2018-07-09
Problem with Filter in Search


uskka:
Description: Filter bar is too close to the footer while loading which looks like something broke. 

Where: Cnx/devb. 
https://cnx.org/search?q=subject:%22Arts%22

![screenshot-cnx org-2018 07 09-15-15-59](https://user-images.githubusercontent.com/40228918/42453664-b855be92-838d-11e8-8137-71084f8e3791.png)

helenemccarron:
@openstaxalina is this within AW's scope?

openstaxalina:
@helenemccarron this is not within AW's scope, but I will keep this in our general backlog.

PiotrKozlowski:
Should we make this look like this while loading?
![image](https://user-images.githubusercontent.com/31478212/43314610-b861e366-9193-11e8-9f51-2177f0ebcc89.png)

openstaxalina:
@PiotrKozlowski are you planning to work on this for the current release (petty.pegasus)?  It would be helpful to discuss the issues that you are working on on the Thursday call.

PiotrKozlowski:
@openstaxalina I already created PR with screenshot how it looks after changes. Of course we can still talk about it.

openstaxalina:
Sliding to q.q. milestone since we did not get to discuss this before code freeze for p.p cc @PiotrKozlowski

PiotrKozlowski:
@Stackblocks , @kerwinso This is already merged to #1816 (petty.pegasus). Can you test it thoroughly during regression testing?

Plan:
When searching for something (normal or advanced search):
  - progress bar should be centered
  - left, grey column should not be visable
Screenshot: https://github.com/Connexions/webview/issues/1703#issuecomment-408371575

openstaxalina:
Thanks, @PiotrKozlowski - I moved this back to petty.pegasus.  cc @helenemccarron

Stackblocks:
In the future we should be sure to try and plan things so we can make sure the testing is completed during the functional verification window, because feature testing during a regression testing is bad practice because chews up regression testing time by having to redeploy one or more times.

Stackblocks:
Displaying as specified with updated version of content / code.

![2018-08-16_15-02-24.gif](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/da052cba-efa9-4058-a316-da54387cd59d)

PR #1805, opened 2018-08-06
issue 1703 add aria-busy attribute to search view and modify design when searching

PiotrKozlowski:
#1703 
Hide filter column when loading and center progress bar. Also add some `margin-top` for header `Wyniki wyszukwiania` > `Search results`
loading

### DIFF
--- a/src/scripts/modules/search/search-template.html
+++ b/src/scripts/modules/search/search-template.html
@@ -1,1 +1,1 @@
-<div class="search"></div>
+<div class="search" aria-busy="true"></div>
--- a/src/scripts/modules/search/search-template.html
+++ b/src/scripts/modules/search/search-template.html
@@ -1,1 +1,1 @@
-<div class="search" aria-busy="true"></div>
+<div class="search"></div>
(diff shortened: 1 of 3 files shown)


src/scripts/modules/search/search-template.html
```diff
@@ -1 +1 @@
-<div class="search"></div>
+<div class="search" aria-busy="true"></div>
```
philschatz: If the AJAX request never starts (because of a bug in the JS code) then it seems like this indicator is incorrect. Could it instead be added when `@model = searchResults.config().load({query: location.search})` is called?

Accurately adding the attribute when the AJAX request is sent may help with testing (to find bugs when the search never began)

src/scripts/modules/search/search.less
```diff
@@ -2,4 +2,18 @@
 
 .search {
   margin: 4rem 3rem;
+  .results {
+    > header {
+      margin-top: 1rem;
+    }
+  }
+  &[aria-busy="true"] {
```
philschatz: Nice selector!

PR #1808, opened 2018-08-06
WIP: less linter

Adrianltd:
Linter is a mistake-preventing tool for developers. It's still in progress. Not in a milestone, because it's a rather low priority side project.

Todo:
- [ ] pre-commit linter deploy
- [ ] establish what config (linter rules and ordering of properties inside of declarations) we need
- [ ] use bash scripts rather than gulp
- [ ] investigate empty variables


PiotrKozlowski:
What about ordering like this:
```
.selector {
  /* Positioning */
  position: absolute;
  z-index: 10;
  top: 0;
  right: 0;

  /* Display & Box Model */
  display: inline-block;
  overflow: hidden;
  box-sizing: border-box;
  width: 100px;
  height: 100px;
  padding: 10px;
  border: 10px solid #333;
  margin: 10px;

  /* Color */
  background: #000;
  color: #fff
  
  /* Text */
  font-family: sans-serif;
  font-size: 16px;
  line-height: 1.4;
  text-align: right;

  /* Other */
  cursor: pointer;
}
```
I took it from `https://css-tricks.com/poll-results-how-do-you-order-your-css-properties/` but I was acctually doing something very similar every time when I was writing css.

Issue #1777, opened 2018-08-01
Incorrect display of the list


wilkus27:
When A,B,C,D list appears in the pdf, such item displays from the new line:
![list_in_text_pdf](https://user-images.githubusercontent.com/40228854/43516590-4607fb74-9586-11e8-9cc0-914fff4c26e0.JPG)

On webview items are in line:
(https://katalyst01.cnx.org/contents/bDIuMp-w@7.1:3E6YBlds@7.1/Science-Practice-Challenge-Que)

![list_kat](https://user-images.githubusercontent.com/40228854/43516635-7848d176-9586-11e8-8300-432ecfb981cb.JPG)

helenemccarron:
@openstaxalina is this something we need to address? First we cannot target only Biology, so this may have wide implications, second, I think this doesn't look bad.

openstaxalina:
@helenemccarron this display is not good; the block of text is difficult to parse and the letters ("A", "B", etc.) are closer to the _preceding_ content.  Let's brainstorm on this one.

@wilkus27 do you have a rough count of how many items like this you found in AP Biology?

wilkus27:
@openstaxalina I noticed that this issue appears through the book just in module **Science Practice Challenge Questions**. If it is needed I can find all of  the exercises where this problem occurs. Up to the 17th module I found 34 items like this.

helenemccarron:
Apologies, this is indeed a bug. The `p` are inline, they _should_ be after @PatBoj  fix for the problems and solutions containers. @PatBoj do you mind taking a look?

PatBoj:
@helenemccarron First paragraph `p` should be displayed as `inline` while other ones should be displayed as `block`. This is how it looks like after my changes. I've done this on my browser only so it's not fixed yet.
![image](https://user-images.githubusercontent.com/20907906/43887842-3d8fb74e-9bc0-11e8-8d86-037c4e9edbc3.png)

helenemccarron:
Perfect, @PatBoj , that's what I thought! Can you please link to this issue so it gets tested along with your changes?

PatBoj:
Ok @helenemccarron, no problem. This issue #1636 is related to this one.

wilkus27:
Verified on katalyst01:

![list](https://user-images.githubusercontent.com/40228854/43957517-5dd7820a-9ca8-11e8-819e-50ace7133698.JPG)

PR #1819, opened 2018-08-09
Fixed incorrect display of lists

PatBoj:
Issue : #1777 

I've changed display style of the paragraphs in exercises and solutions:
- the first paragraph displays `inline`, what's making number of the exercise to be inline,
- other elements are display as `block` like in pdf.

### **Before:**
image

### **After:**
image

### DIFF
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -210,6 +210,9 @@ h1.example-title .text {
   > :first-child {
     display: inline;
   }
+  > :not(:first-child) {
+    display: block!important;
+  }
 }
 
 // solutions of notes, exercises, and examples have a top-border to separate the solution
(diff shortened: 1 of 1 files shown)


Adrianltd:
Do you have to use `!important` on this rule?
Why the first words in paragraphs are bold? Shouldn't the subcaptions be bold instead?

PiotrKozlowski:
Other paragraphs are `inline` because of line 373
```
.os-eoc .exercise {
  p { display: inline; }
...
```
If we will apply these changes, will it cause a regression? Or this is specific for these containers?
```
.os-problem-container, .os-solution-container {
...
> :not(:first-child) {
    display: block!important;
  }
```

helenemccarron:
That's what I thought, you should test it in a couple different books but I believe the inline `p` were introduced to fix the issue you've working on ( `.os-problem-container, .os-solution-container`) and should be ok to remove.

PatBoj:
Thanks @PiotrKozlowski, the code looks better now and works as good as it should be.

Issue #1843, opened 2018-08-16
Bullets in List in Table Cut Off 


openstaxalina:
Webview on the left; PDF on the right:
![bulletscutoff](https://user-images.githubusercontent.com/7863733/44233857-a0bac480-a16a-11e8-980d-7329f6664579.png)

Link to content: https://content02.cnx.org/contents/o0xnhd_t@10.20:DfYAJShE@4/43-Corporations-Limiting-Your-

openstaxalina:
Another example of this content: https://content02.cnx.org/contents/o0xnhd_t@10.20:FeqYe8JF@4/51-Entrepreneurship-Today

lwtchu:
Another example: https://content02.cnx.org/contents/o0xnhd_t@10.20:MwVrrGEx@4/61-The-Role-of-Management#fs-idm226416352

openstaxalina:
Another example: https://content02.cnx.org/contents/o0xnhd_t@10.20:mnLvpdmD@2/53-Small-Business-Driving-Amer

PiotrKozlowski:
Related to #1831

openstaxalina:
Leaving this open for Alan's tracking purposes (more content to test against #1821).

Stackblocks:
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/1d4bb105-7140-4f19-b1b4-cf8b22b16603)

Issue #1840, opened 2018-08-16
Part of the table is not visible 


uskka:
Description: Screen from chrome, when the screen is zoomed in enough dots are visible again.

https://katalyst01.cnx.org/contents/vEmOH-_p@4.48:BoszusFu@4/251-Aggregate-Demand-in-Keynes

![screenshot-katalyst01 cnx org-2018 08 16-12-29-56](https://user-images.githubusercontent.com/40228918/44203745-6d6b3c00-a150-11e8-97c0-b08e6df9d766.png)

openstaxalina:
@uskka could you attach a screenshot of how this looks in the PDF?

PiotrKozlowski:
It's related to #1831

uskka:
Sure, here is the screenshot from the PDF. 

<img width="835" alt="screen shot 2018-08-17 at 10 01 46" src="https://user-images.githubusercontent.com/40228918/44255161-63efdb80-a205-11e8-8010-b3db5ed33240.png">

PiotrKozlowski:
Ready for testing

Issue #1831, opened 2018-08-14
Table with dotted lists don't have margin


wilkus27:
There is no margin in table in which the cell content are dotted lists. Other tables (with text or equations) display correctly.

katalyst01 (https://katalyst01.cnx.org/contents/kRfPjKij@4.5:BoszusFu@4/111-Aggregate-Demand-in-Keynes)

![margin_kat](https://user-images.githubusercontent.com/40228854/44085263-3b7c3a0c-9fb9-11e8-94f3-1dfdb0cda752.JPG)

This is how it looks in pdf:

![margin_pdf](https://user-images.githubusercontent.com/40228854/44085276-480f9a7a-9fb9-11e8-8bc6-6d94875b0fc0.JPG)

wilkus27:
Verified on katalyst01:

![list_table](https://user-images.githubusercontent.com/40228854/44405278-25d21f00-a559-11e8-8108-d9888d8670fd.JPG)

PR #1786, opened 2018-08-02
move css to content-baked

PiotrKozlowski:
fix for requested changes

PR #1845, opened 2018-08-17
issue 1831 / 1840 / 1843 - no padding left for lists inside tables

PiotrKozlowski:
#1831, #1840 and #1843 

I removed old code from `content-common.less`. It was moved to `content-baked`. I also set padding left for all lists inside tables to be 1.5em

table

### DIFF
--- a/src/scripts/modules/media/body/content-common.less
+++ b/src/scripts/modules/media/body/content-common.less
@@ -346,16 +346,6 @@ table {
   background-color: @table-bg;
   counter-increment: table;
 
-  ul {
-    padding: 0;
-    margin: 0;
-    &[data-bullet-style="none"] {
-      li {
-        list-style: none;
-      }
-    }
-  }
-
   caption {
     margin-top: 0.5rem;
     font-size: 1.3rem;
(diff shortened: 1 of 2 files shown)


src/scripts/modules/media/body/content-common.less
```diff
@@ -346,16 +346,6 @@ table {
   background-color: @table-bg;
   counter-increment: table;
 
-  ul {
```
helenemccarron: I am concerned with regression testing if we modify `content-common.less` since this affects every single book that isn't baked (that is a vast majority of our content). I think it might be wiser to overwrite it in `baked-content.less` to avoid a huge regression testing situation.

PiotrKozlowski: @helenemccarron I added this code earlier and you advise me to move it to content.baked so I did but I removed only one part and left behind that part

helenemccarron: Oh, that was added in a different PR? Can you please link it here since it's not obvious in this PR that we are removing something we previously added.

PiotrKozlowski: I added this here `https://github.com/Connexions/webview/pull/1745/commits` and then moved this `https://github.com/Connexions/webview/pull/1786/commits/34c68fe9557bd41e07374207b7fc586fc00c8732` but content is still there

Issue #1835, opened 2018-08-15
Update techsupport@cnx.org email to support@openstax.org


brittweinstein:
#Description
Yesterday, I found the email techsupport@cnx.org as an avenue for support on a couple CNX pages.  I sent an email to this address but do not know where it goes.  It hasn't come to the service desk yet.

I found the techsupport@cnx.org email address here:
https://legacy.cnx.org/help/techsupport under request assistance heading. 
https://legacy.cnx.org/aboutus/contact under technical support
https://legacy.cnx.org/aboutus under Getting Involved

#Impact
* **Severity**: [X ] High [ ] Medium  [ ] Low
* **Frequency of Occurence**: [ ] High [ ] Medium  [ ] Low

#Acceptance
- At https://legacy.cnx.org/help/techsupport under request assistance heading, users are directed to support@openstax.org, which opens a message in the mail client.
- At https://legacy.cnx.org/aboutus/contact under request assistance heading, users are directed to support@openstax.org, which opens a message in the mail client.
- At https://legacy.cnx.org/aboutus under request assistance heading, users are directed to support@openstax.org, which opens a message in the mail client.

TomWoodward:
changes preview on:
https://legacy-staging.cnx.org/help/techsupport
https://legacy-staging.cnx.org/aboutus

let me know when to copy to prod

openstaxalina:
@TomWoodward This looks good; please copy to prod when convenient.  We may want to avoid end of day, just in case anything goes wrong :-)

TomWoodward:
changes made on prod

PR #1851, opened 2018-08-20
update copy on about/contact

TomWoodward:
I used google translate for the polish....

image


PiotrKozlowski:
```
about-contact-questions-content = Możesz skontaktować się z nami wysyłając e-mail na adres <a href="mailto:support@openstax.org">support@openstax.org</a>.
```
We are not sure if polish support have a different email address?

openstaxalina:
@PiotrKozlowski you are correct - Support in Polish should have a different email address.  Do you have a support email address that you would like us to use?

openstaxalina:
@TomWoodward To keep this moving forward, we should use `kontakt@openstax.pl` as the Polish support address.  Katalyst's project manager (who would have the final say) is currently out of office.  However, if she requests any changes, we will know before code freeze next week.  Please let me know if you have any concerns.

cc @PiotrKozlowski

reedstrm:
@openstaxalina recall that this is a Through The Web (TTW) change, so doesn't _have_ to sync up with the release, but it can if we want to do it that way. I checked w/ Ed on what BIT does, he said they sync changes that are required for/related to new features, but that MARCOM makes other changes on their own schedule.

openstaxalina:
@reedstrm Good reminder!  In fact we already released changes for other legacy pages (#1835) off-cycle this morning.  However, I am not comfortable directing traffic to `kontakt@openstax.pl` without Paulina's go-ahead.  As soon as I have word from her, I'll update here and we can proceed as quickly as Tom is available :-)

openstaxalina:
@TomWoodward I got the go-ahead from Katalyst to use the email address - no changes needed. I suggest we could deploy to staging next week, I'll review, then we can deploy to prod.
cc @reedstrm @kerwinso

PR #1854, opened 2018-08-20
clickable search button in book view

TomWoodward:
clickable search

### DIFF
--- a/src/scripts/modules/media/nav/nav.coffee
+++ b/src/scripts/modules/media/nav/nav.coffee
@@ -57,6 +57,7 @@ define (require) ->
       'click .back-to-top > a': 'backToTop'
       'keydown .searchbar input': 'handleSearchInput'
       'click .searchbar > .clear-search': 'clearSearch'
+      'click .searchbar > .fa-search': 'handleSearch'
 
     closeAllContainers: (nodes = @model.get('contents')?.models) =>
       if nodes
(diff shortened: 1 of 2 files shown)


src/scripts/modules/media/nav/nav.coffee
```diff
@@ -147,28 +148,31 @@ define (require) ->
...
+      # before the search has loaded
+      @enableLoadSearch()
+      # after the search has completed
+      BookSearchResults.fetch(options).done((data) =>
+        if not @tocIsOpen
+          @toggleContents()
+        @model.set('searchResults', data.results)
+        @enableClearSearch()
+      ).fail((err) ->
+        console.error("Search failed:", err)
```
philschatz: ooh, is there a quick way to let the user know the search failed? It's not a regression since the original code silently failed as well.

src/scripts/modules/media/nav/nav.coffee
```diff
@@ -57,6 +57,7 @@ define (require) ->
       'click .back-to-top > a': 'backToTop'
       'keydown .searchbar input': 'handleSearchInput'
       'click .searchbar > .clear-search': 'clearSearch'
+      'click .searchbar > .fa-search': 'handleSearch'
```
philschatz: Do other parts of the codebase directly refer to the `fa-` (font-awesome) icon classes? I thought other code relied on an additional class name when targeting them so that we could replace/upgrade the UI (font-awesome) without having to search through references to `fa-` in the JavaScript

TomWoodward: @philschatz in this class in particular there are a lot of references to `fa` and `fa-` classes, not sure about the rest of the project though.

Issue #1855, opened 2018-08-21
Lists in exercises without margins


PatBoj:
That issue was originally created on `cnx-recipes` repo https://github.com/Connexions/cnx-recipes/issues/555. 

Natalia writes:
> After activating _exercise-container_ flag, when the exercise begins with the list, the list displays without margins:
> 
> katalyst01 (https://katalyst01.cnx.org/contents/kRfPjKij@4.5:pKpFhGyF@4.5/Answer-Key)
> 
> ![answer_lists](https://user-images.githubusercontent.com/40228854/44210671-64d23000-a167-11e8-8055-55abf682766e.JPG)

PatBoj:
I've simply changed display style for lists which are direcly below the problem (or solution) number. I've also deleted top margin to those lists in order to reduce the space between the number and first element of the list.

## ***Before:***
![image](https://user-images.githubusercontent.com/20907906/44396373-fa413b80-a53c-11e8-8353-27208f48f68b.png)

## ***After:***
![image](https://user-images.githubusercontent.com/20907906/44396379-ff9e8600-a53c-11e8-9a09-78689f7b7194.png)

## ***PDF:***
![image](https://user-images.githubusercontent.com/20907906/44396382-03320d00-a53d-11e8-8b89-0f37e8f33087.png)

wilkus27:
After deployment in Introductory Statistics text displayed under the number and in Macroeconomics for AP I didn't see any changes. So I tried to bake both of this books. Than in Macroeconomics still were no changes  but Introductory Statistics displayed like before deployment. Next I deployed again and now in both books exercises display incorrectly, e.g.

Introductory statistics(https://katalyst01.cnx.org/contents/MBiUQmmY@20.1:xvUdEJ3p@20.1/Practice):

![stat](https://user-images.githubusercontent.com/40228854/44668939-0e92a600-aa1f-11e8-8f89-da84ec370fdf.JPG)

Macroeconomics for AP(https://katalyst01.cnx.org/contents/kRfPjKij@4.5:cW9mg8jr@4.5/Review-Questions)

![macro](https://user-images.githubusercontent.com/40228854/44669034-3c77ea80-aa1f-11e8-97e7-f2d52e0d1944.JPG)

helenemccarron:
[...]

PR #1857, opened 2018-08-21
Fixed lists' margin in exercises

PatBoj:
Issue #1855, but originally it was created on `cnx-recipes` repo https://github.com/Connexions/cnx-recipes/issues/555.

I've simply changed display style for lists which are direcly below the problem (or solution) number. I've also deleted top margin to those lists in order to reduce the space between the number and first element of the list.

## ***Before:***
image

## ***After:***
image

## ***PDF:***
image

### DIFF
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -207,8 +207,11 @@ h1.example-title .text {
 .os-problem-container,
 .os-solution-container {
   display: inline;
-  > :first-child {
+  > :first-child:not(ul):not(ol) {
     display: inline;
+  }
+  > ul, > ol {
+    margin-top: 0rem;
   }
 }
 


src/scripts/modules/media/body/content-baked.less
```diff
@@ -207,9 +207,12 @@ h1.example-title .text {
 .os-problem-container,
 .os-solution-container {
   display: inline;
-  > :first-child {
+  > :first-child:not(ul):not(ol) {
```
Adrianltd: `> :first-child:not(ul):not(ol)` looks strange to me.
It works for the first child of any selector inside the `os-problem-container` and `os-solution-container` that is not a list. Should it works for those in all of the books?

Adrianltd: That's actually connected with https://github.com/Connexions/cnx-recipes/issues/571 which will be solved by https://github.com/Connexions/webview/pull/1867
But it's also good to prevent this behaviour for lists in `os-eoc` and `os-eob`.

PR #1867, opened 2018-08-23
fixed numbering and display of exercises

Adrianltd:
Numbering fixed in recipes https://github.com/Connexions/cnx-recipes/pull/565 and display is fixed in this PR.
Basically I've changed this commit https://github.com/Connexions/webview/commit/a1f00679b7910be449a2c7dc90b89e4577c5c461 to be more specific (in the `os-eoc` and in the `os-eob`, is that correct @PatBoj ?)

### DIFF
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -204,11 +204,13 @@ h1.example-title .text {
   padding: 0;
 }
 
-.os-problem-container,
-.os-solution-container {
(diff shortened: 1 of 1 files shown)


Adrianltd:
This should be also fixing https://github.com/Connexions/cnx-recipes/issues/571

helenemccarron:
This might conflict with @PatBoj PR (#1862 ), are they both necessary. should they be combined?

PR #1875, opened 2018-08-27
use browser download for 'download for free' button

TomWoodward:
(Alternative to #1876)


kerwinso:
We went with #1876 instead.

PR #1876, opened 2018-08-27
let the browser Download PDFs

philschatz:
this was causing Safari & Edge browsers to fail

An untested attempt at a fix.

(Alternative to #1875)

### DIFF
--- a/src/scripts/loader.coffee
+++ b/src/scripts/loader.coffee
@@ -62,9 +62,8 @@ define (require) ->
       # Only handle links intended to be processed by Backbone
       if e.isDefaultPrevented() or href.charAt(0) is '#' or mailto.test(href) then return
 
-      e.preventDefault()
-
       if external.test(href)
+        e.preventDefault()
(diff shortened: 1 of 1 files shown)


src/scripts/loader.coffee
```diff
@@ -62,9 +62,8 @@ define (require) ->
       # Only handle links intended to be processed by Backbone
       if e.isDefaultPrevented() or href.charAt(0) is '#' or mailto.test(href) then return
 
-      e.preventDefault()
-
       if external.test(href)
+        e.preventDefault()
```
TomWoodward: but you still go through here to get to line 76

src/scripts/loader.coffee
```diff
@@ -74,11 +73,13 @@ define (require) ->
           content = href.match(exports)
           router.navigate("/donate/download/#{content[1]}/#{content[2]}", {trigger: true})
```
TomWoodward: you probably still want a prevent default here right?

PR #1880, opened 2018-08-29
Small display change exercises properly

PatBoj:
Small fix to repair some issues.


PatBoj:
@Adrianltd It is related to this issue https://github.com/Connexions/webview/issues/1860. I will double check if it is necessary.

PatBoj:
@Adrianltd You were right, I've deleted those selectors.

Issue #1883, opened 2018-08-29
Screen is scrolled to the bottom of the page when history button is clicked 


uskka:
Screen is scrolled to the bottom of the page when history button is clicked on Google Chrome and Opera while when the same button is clicked on Firefox screen is in the same place where it was. I think that the first behaviour is an issue. 

https://katalyst01.cnx.org/contents/tWu56V64@34.15:fVN7UicZ@7/7-4-Finite-Population-Correction-Factor

**Firefox**
![screenshot-katalyst01 cnx org-2018 08 29-15-58-03](https://user-images.githubusercontent.com/40228918/44793115-cbaf0a80-aba5-11e8-8db3-7a6d648bf399.png)

**Chrome**
![0](https://user-images.githubusercontent.com/40228918/44793127-d2d61880-aba5-11e8-8c35-a5c469dfac05.png)

openstaxalina:
Agree that the *desired behavior* is that it is scrolled to the *top* of the History section.  We need to change the behavior in Chrome and Opera.  We need to check Safari.

uskka:
On Safari it works fine.

PiotrKozlowski:
Ready for testing

uskka:
Works fine. 
![screenshot-katalyst01 cnx org-2018 09 11-16-08-40](https://user-images.githubusercontent.com/40228918/45365826-b098bd80-b5dd-11e8-9e39-099ce050819b.png)

PR #1893, opened 2018-09-03
issue 1883 stay at current scroll position after opening history tab

PiotrKozlowski:
#1883 
Check if `scrollY` has changed after opening tab and if so then return to starting position.


philschatz:
@TomWoodward recently ran into scrolling issues so I added him as a reviewer.

PR #1894, opened 2018-09-04
fix image height in about this book section

TomWoodward:
setting image height automatically in flexy containers is janky

### DIFF
--- a/src/scripts/modules/media/body/body.less
+++ b/src/scripts/modules/media/body/body.less
@@ -65,7 +65,7 @@
 
     .media-body-about-cover {
       width: ~"calc(125px + 2rem)";
-      height: intrinsic;
+      height: ~"calc(125px + 2rem)";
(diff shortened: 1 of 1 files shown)


kerwinso:
See #1895 for more details on the bug.

src/scripts/modules/media/body/body.less
```diff
@@ -65,7 +65,7 @@
 
     .media-body-about-cover {
       width: ~"calc(125px + 2rem)";
-      height: intrinsic;
+      height: ~"calc(125px + 2rem)";
```
philschatz: Could the height just be omitted, rather than forcing the image to be a square?

TomWoodward: no, it ends up all kinds of janky heights because the browser sometimes sets its height based on the un-flexy width of the image

Issue #1897, opened 2018-09-04
Baked Books No Longer Landing on First Intro Module


openstaxalina:
We have some code in webview so that when a user navigates to a book, they land on the first module after the preface, aka the first introduction module.

Demonstrated here or with any other unbaked book on cnx.org: https://cnx.org/contents/jVCgr5SL 

On staging.cnx.org, users are landing on the preface, as demonstrated here: https://staging.cnx.org/contents/jVCgr5SL.

# Acceptance Criteria
- [ ] when a user navigates to a book via the book ID (such as jVCgr5SL or any OpenStax book clicked from the homepage), the user lands on the first module after the preface, regardless of whether the book is baked.
- [ ] that page has a cover and about text, plus a hyperlinked ToC as coded here: https://github.com/Connexions/webview/pull/1886

PiotrKozlowski:
Ready for testing.

wilkus27:
Verified on katalyst01. 
After clicked on any  book (baked or unbaked) from the homepage user landing on the first module after Preface.

unbaked book:

![prealgebra](https://user-images.githubusercontent.com/40228854/45416260-d080bd80-b67f-11e8-8eaa-268df057d49a.JPG)

baked book:

![concepts](https://user-images.githubusercontent.com/40228854/45416289-e4c4ba80-b67f-11e8-9a4c-651eeff0c20c.JPG)

In parts of books there is a problem with displaying of a cover, but this issue is already resolved on staging.cnx.org by https://github.com/Connexions/webview/pull/1894 and https://github.com/Connexions/webview/issues/1895.

wilkus27:
@helenemccarron @PiotrKozlowski After today deployment only unbaked books display correctly. Clicking on baked book from Homepage user is landing in Preface again.

PiotrKozlowski:
It's cache or deployment issue. My code is not on the server.
cc @wilkus27 @helenemccarron

helenemccarron:
I redeployed and refreshed the cache and believe it is now working as intended. @wilkus27 can you confirm: https://katalyst01.cnx.org/?

PiotrKozlowski:
@helenemccarron Your deploy took only 3 minutes and I can't see new files on katalyst server inside `/var/www`, probably server is again out of space.

kerwinso:
Addressed in #1953.

kerwinso:
[...]

PR #1899, opened 2018-09-05
issue 1897 change regex to match also baked books

PiotrKozlowski:
#1897 
`@getPage(result).get('title')` for `unbaked` books was returning `Preface`
for `baked` books: `<span class="os-text">Preface</span>`

Regex before change: `match(/^Preface/)` `^` is checking if text starts with `Preface`
Regex after change: `match(/Preface/g)` is checking if text contains `Preface`

### DIFF
--- a/src/scripts/models/content.coffee
+++ b/src/scripts/models/content.coffee
@@ -50,7 +50,7 @@ define (require) ->
     defaultPage: ->
       result = 1
       if @isBook() and @isCcap()
-        result += 1 while @getPage(result).get('title').match(/^Preface/)
+        result += 1 while @getPage(result).get('title').match(/Preface/g)
       return result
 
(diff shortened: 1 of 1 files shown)


src/scripts/models/content.coffee
```diff
@@ -50,7 +50,7 @@ define (require) ->
     defaultPage: ->
       result = 1
       if @isBook() and @isCcap()
-        result += 1 while @getPage(result).get('title').match(/^Preface/)
+        result += 1 while @getPage(result).get('title').match(/Preface/g)
```
Adrianltd: You can remove the `g` from the regex `/Preface/g`. This modifier makes a global search, so it finds all matches. Without it, the search stops at first match.
It shouldn't make a difference though, because each book has only one Preface.

Issue #1892, opened 2018-08-30
Remove donation prompts for CNX Downloads


openstaxalina:
We should remove the donation prompts when a user wants a file (PDF, EPUB, or ZIP) from cnx.org.

The pages are confusing to users.  Sometimes users think they must donate to receive a PDF.  CS reports that this causes confusion about our mission, and frustration as well as requests for refunds.  Also, if the download button ever malfunctions as it did in last week's critical situation, users are even more convinced they must donate...a mandatory donation isn't really a donation!  

# Acceptance Criteria
- [x] when a user selects an option under the "Get This Book!" button toward the top right of a page in a book on cnx.org, they are **not** directed to a donation prompt.  Instead, the download is initiated.
- [x] when a user selects an option under the "Downloads" tab on the bottom of a page in a book on cnx.org, they are **not** directed to a donation prompt.  Instead, the download is initiated.
- [ ] add automated tests to ensure the correct link is present 

# Example of a donation prompt page:
<img width="1421" alt="screen shot 2018-08-30 at 3 44 42 pm" src="https://user-images.githubusercontent.com/7863733/44878118-aab0ed00-ac6b-11e8-9abe-f4374e58204f.png">

# PRs
https://github.com/Connexions/webview/pull/1907

kerwinso:
Verified on qa.cnx.org, webview sha `7f2345288efd81088783b66b5beebc5748a5c5bb`.

kerwinso:
Verified working on staging.

Issue #1891, opened 2018-08-30
Redirect CNX Give page to OpenStax Give page


openstaxalina:
We should redirect the “Give” page on CNX.org to the OpenStax “Give” page.  This has several benefits:
(a) users will be added to Salesforce (and all the benefits, both to users & OpenStax, that come with that)
(b) users will receive a properly-formatted gift receipt and have the opportunity to learn more about OpenStax
(c) easier for accounting purposes; the funds are not separate

# Acceptance Criteria
- [x] link to https://openstax.org/give from the cnx.org header
- [x] https://cnx.org/donate redirects to https://openstax.org/give 
- [x] when a user clicks "Give" from the CNX homepage, they are redirected to **in the same tab**
- [ ] regression tests are added/updated to cover the test cases above

# PRs
https://github.com/Connexions/webview/pull/1907

kerwinso:
Verified working on qa.cnx.org, webview sha `7f2345288efd81088783b66b5beebc5748a5c5bb`.

@TomWoodward if you want to automate the regression test, it's here: https://openstax.testrail.com/index.php?/cases/view/210412
Slack me and I can get you a Testrail login.

kerwinso:
Verified working on staging.cnx.org.

PR #1907, opened 2018-09-05
remove donate pages and redirect to openstax/give

TomWoodward:
#1891
#1892


src/scripts/router.coffee
```diff
@@ -52,8 +52,7 @@ define (require) ->
         # Sending analytics will be done once the path is canonicalized
 
       @route /^donate\/?([^/\?;]*)?\/?([^/\?;]*)?\/?([^/\?;]*)?(?:\?)?.*/, 'donate', (page, uuid, type) ->
-        @appView.render('donate', {page: page, uuid: uuid, type: type})
-        analytics.sendAnalytics()
+        window.location = 'https://openstax.org/give'
```
philschatz: Is this just to handle the case when someone lands directly on the Donate Page (e.g. from a bookmark or from a search engine)? If so, nice!

Issue #1829, opened 2018-08-13
Extra line with license info appears near footer when using Back/Next buttons


kerwinso:
## Steps to Reproduce
1. In any browser, go to any book in CNX e.g., https://cnx.org/contents/CImQfPDv@3.11:GfaWl1GG@4/Introduction
1. Click Next to go to the next page.
1. Scroll to the bottom of the page.
1. Notice a line right below the 'Downloads |History |Attribution |More Info' tabs that says, `This work is licensed by Rice University under a Creative Commons Attribution License (by 4.0).`

This line disappears if you reload the page. But it shouldn't display at all.

![image](https://user-images.githubusercontent.com/11298165/44056464-f07803c0-9f0d-11e8-9182-37525b2bb012.png)

# PR
#1932, merged into #1925

openstaxalina:
It also appears when you navigate via clicking in the Table of Contents.  When the contents of the book content area is replaced, rather than the entire page being rendered again, the license duplication occurs.

brittweinstein:
@PiotrKozlowski assigning to you for work in down time :) If you have any questions, please message @TomWoodward as he may be working on something similar.

PiotrKozlowski:
I've merged changes into #1925 It's ready for testing

kerwinso:
Verified with webview sha `5a3df2498f2b5037fe4d051e839b4d8d75580322` on qa.cnx.org -- the offending line has disappeared, with no noticeable regressions. Will test again (with full regression testing) once this code is deployed to staging.

kerwinso:
Line all gone on staging

PR #1932, opened 2018-09-11
issue 1829 - disable license view

PiotrKozlowski:
#1829 
disable license view

I've commented those lines instead of deleting them because I'm not sure if this will be needed in future


Adrianltd:
I think you can delete those lines, they will stay in GIT history anyway.

Issue #1882, opened 2018-08-29
Buttons on the bottom of the page move when clicked


uskka:
All of the buttons on the bottom of the page move slightly when clicked. Appears in all of the books.

Reproduction steps:

1.Scroll to the bottom of the page.
2. Click any button. 
3. Click on the background. 

https://katalyst01.cnx.org/contents/tWu56V64@34.15:fVN7UicZ@7/7-4-Finite-Population-Correction-Factor

![screenshot-katalyst01 cnx org-2018 08 29-15-40-53](https://user-images.githubusercontent.com/40228918/44791468-f72ff600-aba1-11e8-9403-27e801030696.png)

openstaxalina:
@uskka Are you able to capture a .gif or video screen capture of this issue?  Are there any patterns that you can identify as to when this occurs?

uskka:
This is how it looks like. 3 screens one for every reproduction step. It occurs every time I try to reproduce it.

1.  Scroll to the bottom of the page.
2. Click any button.
3. Click on the background.

![output_yfkjbs](https://user-images.githubusercontent.com/40228918/45618993-fccd7d00-ba77-11e8-929e-05c8a2efc77e.gif)

PiotrKozlowski:
Ready for testing

uskka:
Fixed. Verified on katalyst01.
![output_ehpvnk](https://user-images.githubusercontent.com/40228918/46008403-ee591400-c0bc-11e8-9e27-7dfb97c10d77.gif)

PR #1950, opened 2018-09-19
issue 1882 remove border on focus

PiotrKozlowski:
#1882
There was additional `border` on `:focus` which was causing moving all table. This border is not needed because other elements does not have border so when it's added position of other elements have to be updated.

### DIFF
--- a/src/scripts/modules/media/footer/footer.less
+++ b/src/scripts/modules/media/footer/footer.less
@@ -39,7 +39,8 @@
       }
 
       &:focus {
-        outline: 0;
+        outline: 1px;
+        outline-style: solid;
       }
 
       &.active,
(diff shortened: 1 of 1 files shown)


src/scripts/modules/media/footer/footer.less
```diff
@@ -39,9 +39,6 @@
       }
 
       &:focus {
-        border-top: 0.1rem solid @brand-primary;
-        border-left: 0.1rem solid @brand-primary;
-        border-right: 0.1rem solid @brand-primary;
         outline: 0;
```
Adrianltd: [accessibility]
When you use `tab` key to navigate through the website, you should be able to see focus on elements. 
Those rules which you have removed actually disable the default focus visibility and add borders instead. 

My suggestion is to remove the 41 and 42 lines too and there will be default focus which shouldn't affect the layout.

PiotrKozlowski: I've changed to:
```
outline: 1px;
outline-style: solid;
```
This is giving same result as setting `border` but will not cause moving elements.

PR #1952, opened 2018-09-19
Less linter

Adrianltd:
This is a new PR with recreated linter https://github.com/Connexions/webview/pull/1808. What has changed is:
- Linter is now working directly using CLI instead of task runner
- Linter is auto deployed after `git commit`. It lints all staged *.less files, if build passes it adds them and commits, if fails - it stops with no commit. (pre-commit using git hooks)
- I have established a config that seemed reasonable for me at that moment, now we need to investigate which rules are really needed.

- [x] Make it install with one command.
- [ ] Investigate empty variables errors in Travis
- [x] Lint ordering of attributes
- [x] Establish best config

**Install and run instructions**
Install:
1. `git checkout huskylinter`
2. `npm install`

Run:
1. Make changes in any `*.less` file in the repo.
2. `git add path/to/the/changed/file`
3. `git commit -m "commit message"`
4. Linter should run:
a) if the build **passes**, files are successfully commited
b) if the build **fails**, no files are commited and you can see all errors and warnings in your terminal


helenemccarron:
⚠️ Wall of text below ⚠️ 

### Linter documentation used in this PR:
https://stylelint.io/user-guide/rules
https://stylelint.io/user-guide/about-rules 

### Log Analysis
I ran into an `npm install` issue but I was able to run the linter on my staged file. I used `scripts/modules/media/body/content-baked.less` in its current state as my test data.

During my first pass at reading the log, here are my thoughts. Please note this was before I read through the list of rules, so this is really my first feel for the log without looking at how it was set up:

- `✖    9:3   Expected single space after ,` and `✖    9:4   Unexpected whitespace at end of line` I think it doesn’t like adding each selector on a new line, which we could add as a rule (edited)

- `⚠  365:19  Unexpected leading zero` seems a bit weird to me, I’ve never considered using .2 instead of 0.2, if you want we can bring that to the issue do discuss (edited)

- `✖   92:16  Unexpected unit rem for property font-size`  I guess we need to add `rem` to the whitelist, as well as `px` or `em` or decide on which units we are whitelisting and blacklisting. I think right now, it's a mix of all of them.

- `✖  123:38  Unexpected empty block` I am not sure i understand what this is about. Additionally it doesn’t seem to complain about the indentation on this line.

- `✖  267:3   Expected selector .os-note-body img to come before selector .splash:not([data-orient="vertical]) img"` non-descending specificity seems a bit jacked up, their suggestion is never accurate, maybe we should disable it. (Note: reading the doc, I understood better what it does but as is, it was pretty confusing. In essence, it's preventing less specific css to be defined after more specific rules, which I think is good. Their suggestions, as mentioned, aren't really useful though. We might need to tweak it).

- `✖  351:1   Unexpected duplicate selector ".os-eoc, .os-eob", first used at line 207` is **really** cool! (and we should fix it too :smile: ). Same for `✖  462:5   Unexpected duplicate display`


***Interesting fact***: I ran a linter plugin in Atom (beautify) on the same file and ran it through the linter a second time. I get roughly the same amount of errors. I believe that disabling some of the rules I mentioned about would help align them a little more (I am not saying Beautify format is what we should aim for but it's in my opinion pretty close to our current practices).

### Possible rule additions
After going over the majority of the rules listed in https://stylelint.io/user-guide/about-rules, here is what I think we could add (I confess I didn’t double check if some of those were already in there). I might have missed some that might be useful but the list below is what jumped at me:

- `indentations` from what I read, as `int`, with a value of `2`
- `max-empty-lines: 1` to avoid wasted vertical space
- `custom-property-pattern:` lower-case-with-dashes
- `declaration-bang-space-after: always`
- `declaration-empty-line-before: always`
- `selector-combinator-space-after: always`
- `selector-list-comma-newline-after: always` (this should fix `✖    9:3   Expected single space after ,` I think)

package.json
```diff
@@ -37,11 +44,16 @@
     "grunt-nginx": "~0.2.2",
     "grunt-string-replace": "~1.3.1",
     "grunt-targethtml": "~0.2.6",
+    "husky": "^0.14.3",
```
philschatz: Is this package used anywhere? From its description it seems like it needs a configuration block in `package.json`

package.json
```diff
@@ -3,12 +3,19 @@
   "version": "0.23.3",
   "description": "Connexions (cnx.org) web site",
   "scripts": {
+    "precommit": "lint-staged",
```
philschatz: can you include linting into the `posttest` command? That way Travis will run it too! Maybe `"posttest": "lint-staged && node ./script/check-templates-for-non-i18n-text.js",`

package.json
```diff
@@ -3,12 +3,19 @@
...
     "prepublish": "bower install --config.analytics=false",
     "upgrade": "npm update; bower update --config.analytics=false",
     "test": "grunt test --verbose",
     "posttest": "node ./script/check-templates-for-non-i18n-text.js",
     "dist": "grunt dist --verbose"
   },
+  "lint-staged": {
+    "*.less": [
+      "stylelint --syntax=less --custom-formatter=node_modules/stylelint-formatter-pretty",
```
philschatz: Could/should this also include the `--fix` arg?

.stylelintrc.json
```diff
@@ -0,0 +1,85 @@
+{
+  "rules": {
```
philschatz: Is there an existing config that we could extend?

philschatz:
By the way, I snagged this (minus the git hooks) and added it to https://github.com/philschatz/repo-template#readme . Hopefully all of our repos will include similar ways/rules for linting!

One thing to note: in that repo, there is a `npm run-script lint` script that runs different linters for different languages: JavaScript (`lint:js`), TypeScript (`lint:ts`), Markdown (`lint:md`), SASS/CSS (`lint:css`) and shell scripts etc). It might be useful to do something similar in this repo.

Adrianltd:
@helenemccarron:
`✖ 9:3 Expected single space after ,` - yes, this makes you write selectors next to each other and the other rule disallows end of line spaces. We could decide if we want `h1, h2, h3 {` or ```h1,
h2,
h3 {```
`⚠ 365:19 Unexpected leading zero ` - There's actually no other reason for not using leading zeroes (or zeros?:D) than http://codeguide.co/#css-syntax or https://google.github.io/styleguide/htmlcssguide.html#Leading_0s
I was wondering what can be more misunderstanding while scanning the code eg.:
```margin: 0 5em;
margin: 0.5em;```
or 
```margin: 5em;
margin: .5em;```
As you can see, both leading zero and it's absence can be confusing, so we just have to choose one and stick to it.
`✖ 92:16 Unexpected unit rem for property font-size` - I was thinking about choosing which units should be used with which properties, so we could have the best layout which doesn't look like:
https://media.giphy.com/media/13XW2MJE0XCoM0/giphy.gif

We have units like: px, em, rem, %, vw, vh, vmin, vmax... For some people it might be a random choice. Oh, there's also calc(); to get things more complex :face_with_head_bandage: 
I recommend reading those before we make a decision:
https://engageinteractive.co.uk/blog/em-vs-rem-vs-px
https://css-tricks.com/fun-viewport-units

`✖ 123:38 Unexpected empty block` - this basically means that you can't write sth like:
`.class { }`
There always has to be something between `{}`

`✖ 267:3 Expected selector .os-note-body img to come before selector .splash:not([data-orient="vertical]) img"` - I am not sure how it works behind the scenes, but it just shows you that there is a piece of code in line 267 which is overwriten by the other code, so you have to either remove it or move before that other selector. It's just for making our css files shorter. It's hard to maintain, because they both work for img tags in 2 different containers, but it seems that it compares the nesting level. Should we remove this rule?

`custom-property-pattern:` arghh, I'm not the best with regex, anyone wants to help?

`declaration-bang-space-after: always` this will make us do: `.class { attribute: value! important; }` instead of `.class { attribute: value !important; }`

`declaration-empty-line-before: always` do we want that? https://stylelint.io/user-guide/rules/declaration-empty-line-before/

`selector-combinator-space-after: always` and `selector-combinator-space-before: always` are already there, so we would write `element > element` not `element> element` or `element >element`

`selector-list-comma-newline-after: always` how about `selector-list-comma-newline-after: alway-multi-line`? https://stylelint.io/user-guide/rules/selector-list-comma-newline-after/

.stylelintrc.json
Adrianltd: This is the config that I came up with. There are some existing configs like:
https://github.com/stylelint/stylelint-config-standard
https://github.com/stylelint/stylelint-config-recommended
https://github.com/JetBrains/stylelint-config
And lots more, so it's up to you guys to decide if we want to create our own, extend the default or pick someone elses.
I have just found a nice generator for configs https://maximgatilin.github.io/stylelint-config/

package.json
Adrianltd: `npm run precommit` would work for staged file or `stylelint path/to/the/file/you/want/to/lint`
`stylelint path/to/the/file --syntax=less --custom-formatter=node_modules/stylelint-formatter-pretty` would be even better.

package.json
Adrianltd: Yes, this will autofix some of the issues according to the ruleset.

package.json
Adrianltd: Husky is a Git hook manager, it's used to add precommit hook

Adrianltd:
@philschatz linting the files before committing them forces devs to write working and clean code according to the chosen guidelines. It also helps with testing your own code to eliminate incorrect code.
I think that we can include any scipts as pre-commit hooks as long as they will take few seconds at max, because when you do a lot of commits, you will be annoyed with that delay.

It all runs in this order:
1. You change the code.
2. You `git add` the changes.
3. You try to `git commit` them.
4. Precommit hooks are deployed, it the code passes the precommit scripts - it's added and commited. If not, there's no commit added. 

So as long as you add things that work and pass the tests - you can commit those changes. I am not sure what happens when you add few files, and some of them pass tests and others don't. That needs to be checked to fully answer your comment. 
After all that process you can safely undo some commits/squash them etc, but again, but this also needs to be tested if after doing eg. `git rebase i #commit` , squashing commits or cherry-pick, the precommit hooks deploy or not.

Adrianltd:
@philschatz could you take a look on Travis? I think it complains on old version of node.
I have updated the linter with linting of property order and removed `--fix` flag for now, because it created more errors with each lint deploy. I need to investigate why that happens.

philschatz:
It seems like one of the packages added to `package.json` require a specific version of phantomjs which is causing the error. I updated the `.node-version` **and** `.travis.yml` file to be `10.13.0`. That seems to fix the problem but it seems that a couple of tests need to be updated.

Adrianltd:
@philschatz I think it's ready to go. I have only removed `--fix` flag, because for some unknown reason it was adding empty lines in places that it shouldn't. 
During merging someone will need to fix the `less` files, so other people would be able to start working with the correct code-style. I can do it. I can also try to update the tests.
@helenemccarron what do you think? Is the config alright? Feel free to adjust it.

Adrianltd:
We need to run the linter on all the less files that exist right now on the master branch and fix all the issues with it, then we can safely merge.
But we need @helenemccarron approval. Can you please try to lint any bigger `.less` file and see if there's no useless restrictions? If the config is fine with you, I can start fixing all of the files since monday.

philschatz:
I was unable to run the tests locally so was unable to find out why they were failing.

Would it make sense to split this PR into 2? One that adds the linter and another that adds the book description and `aria-label` on the ToC? 

I think the test failures are being caused by the latter

Adrianltd:
@philschatz I have removed all of the files that were previously merged from master, updated the code with the current master.
Locally there are 2 tests failing, on the server there's only 1 failing:
------Server-----
` 1) MediaBodyView tests "before each" hook for "should return the correct url for each exercise":`

------Locally-----
> 1) backbone base view helper tests get template data tests should get template data from models and collections:
>2) MediaBodyView tests "before each" hook for "should return the correct url for each exercise":

Both of those were just fine some time ago.

philschatz:
In that case it looks like the tests no longer reliably pass so they are no longer useful.

As a temporary solution I commented the test so they no longer run.

Jenkins is still failing because the Dockerfile still needs to be switched to use `openstax/node:10.13.0` (see [Slack Discussion](https://openstax.slack.com/archives/C0LA54Q5C/p1549484004677400) for details

brittweinstein:
@helenemccarron  can we close this?

helenemccarron:
I believe @Adrianltd is  still working on this and really close to finishing it.

brittweinstein:
@Adrianltd can this be closed?

Adrianltd:
@brittweinstein This only needs to resolve conflicts with the latest changes on master and switching the Dockerfile to use other node version. It's a small change, but I did not have time to resolve those. I'm hoping to have some spare time after finishing Precalculus.

Adrianltd:
@philschatz I've used `openstax/nodejs:10.13.0`, `openstax/node:10.13.0` and `node:10-jessie` node images and Jenkins still fails. Could you look into the logs (I don't have permission to view Jenkins details) and see if that's still the same thing that fails?

Adrianltd:
@philschatz, @helenemccarron Jenkins builds pass. Big thanks to @pumazi for the latest fix!

Issue #1947, opened 2018-09-18
Figure text wrapping around image


marcjeannott:
**Problem**

The figure text seems to be wrapping around the image on staging.

![text wrapping](https://user-images.githubusercontent.com/33228728/45703670-a117f800-bb3a-11e8-9902-a9166a0e3899.jpg)

This is unlike the other figures found in this book.

**More Detail**

Link - https://staging.cnx.org/contents/jQSmhtXo@19.1:WwXAbgy9@19.1/Answer-Key
This also seems to be appearing with the equation numbering.

Adrianltd:
@marcjeannott 
1. Equations in the Answer Key should be `unnumbered`. Could you create new issue for that and try to spot some of those that are still numbered (we add `unnumbered` class in the recipes, but it seems it doesn't catch all of them).
2. Could you move those issues then to `katAW.sleepy.siren` milestone?

@helenemccarron correct me if I'm wrong

openstaxalina:
This issue is specific to the text wrapping around the figure.  The equation numbering issue (https://github.com/Connexions/cnx-recipes/issues/636) was closed.  Sliding this to the next milestone.

PiotrKozlowski:
Fixed:
![figcaption](https://user-images.githubusercontent.com/31478212/45872038-47f4d200-bd8f-11e8-9580-09e535a29c35.png)

helenemccarron:
@openstaxalina I commented on the PR but do we support unnumbered figures with captions in the Answer Key? This might be why the display of this particular figure looks weird.

openstaxalina:
Issue moved to [Connexions/content-managers #322](https://github.com/Connexions/content-managers/issues/322) via [**ZenHub**](https://www.zenhub.com/)

PR #1955, opened 2018-09-21
issue 1947 style figcaption in eoc and eob

PiotrKozlowski:
#1947 

Now figcaption is matching pdf, but I think it will look better if text is centered. What do you think?

figcaption


helenemccarron:
@PiotrKozlowski, I thought images were left aligned in the EOC and EOB? Cc @PatBoj and @openstaxalina

helenemccarron:
@PiotrKozlowski per Alina, this PR is no longer necessary, she will move the issue to content.

Issue #1951, opened 2018-09-19
Book description text issue on mobile


edwoodward:
The book description in the first Intro page is being pushed far to the right by the book cover. The cover should be above the text on mobile instead of to the left. The problem was seen on Android and iPhone.

![screenshot_20180919-073548](https://user-images.githubusercontent.com/1394695/45756233-ffe67b80-bbe5-11e8-8cf0-4424d9edbdc2.png)

Adrianltd:
@edwoodward could you verify the fix?

openstaxalina:
Per conversation on the call, Stefan (and/or Natalia) will test this.

uskka:
Works fine. Tested on katalyst01.

**IOS**
![42730735_597082417356617_8514398005732835328_n](https://user-images.githubusercontent.com/40228918/46279976-eee52500-c56a-11e8-807d-4a2c3fd6d130.jpg)

** Android**
![screenshot_20180929-081624_samsung internet](https://user-images.githubusercontent.com/40228918/46279923-c0674a00-c56a-11e8-9911-afe41e24a477.jpg)

PR #1958, opened 2018-09-25
1951 mobile book description

Adrianltd:
https://github.com/Connexions/webview/issues/1951

I'm not sure if that's the proper `max-width` to be used. Should I specify this to be `screen` and `portrait` also or it's ok the way it is?
Could you please verify that?

### DIFF
--- a/src/scripts/modules/media/body/body.less
+++ b/src/scripts/modules/media/body/body.less
@@ -79,6 +79,17 @@ body {
   .transition(background 0.5s ease-in-out);
 }
 
+@media (max-width: @screen-xs-max) {
+  .media-body-about {
+    .media-body-about-description-wrapper {
+      display:block;
+      .media-body-about-cover {
+        margin-bottom: 2rem;
+      }
+    }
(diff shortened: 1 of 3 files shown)


src/scripts/modules/media/body/body.less
```diff
@@ -79,6 +79,17 @@ body {
   .transition(background 0.5s ease-in-out);
 }
 
+@media (max-width: @screen-xs-max) {
```
PiotrKozlowski: You can try even simpler trick because it's already using flexbox so you will not have to use `@media`.
Just set `flex-wrap: wrap;` to `.media-body-about-description-wrapper` and `min-width: 400px;` to `.media-body-about-description`. It will automaticall go below cover when hit `min-height`.

You can experiment with `flex: 1;` on `` and move `margin-bottom` from `...wrapper` to `...cover` and so on.

PiotrKozlowski: I'm okay with that

Issue #1757, opened 2018-07-26
a11y: Ensure the solution is read when exercise "Show Solution" is clicked (aria-live?)


brittweinstein:
@brittweinstein commented on [Thu Jul 26 2018](https://github.com/Connexions/content-managers/issues/165)

From https://drive.google.com/file/d/0BxiTbl6K_ZNRelpxRlkwYkZnb3c/view

Issue 10: There is no feedback for 'show solution' buttons
Video example: https://drive.google.com/file/d/0B2fFZC6BqvTLTVJkUXdRUmN1UVk/view?usp=sharing
Reference to standards
1.3.1 Info and Relationships: Information, structure, and relationships conveyed through presentation can be
programmatically determined or are available in text. (Level A)
Suggestion
After selecting the ‘show solution’ button screen reader should read the feedback with the correct answer.

openstaxalina:
@helenemccarron will analyze and assign points.

brittweinstein:
Can slide if needed

helenemccarron:
Addressed in PR #1816

Stackblocks:
With this in place clicking on `[Show Solution]` causes JAWS to read out `ｏｐｅｎ　ｂｒａｃｋｅｔ　ｓｈｏｗ　ｓｏｌｕｔｉｏｎ　ｃｌｏｓｅ　ｂｒａｃｋｅｔ　ｕｓｅ　ｊａｗｓ　ｋｅｙ　ｐｌｕｓ　ｍ　ｔｏ　ｍｏｖｅ　ｔｏ　ｃｏｎｔｒｏｌｌｅｄ　ｅｌｅｍｅｎｔ　ｓｈｏｗ　ｈｉｄｅ　ｓｏｌｕｔｉｏｎ` instead of reading out the solution. Removing from petty.pegasus and sliding.

openstaxalina:
Since this issue will be fixed with baked books, we could defer this change.   In Larissa's terms, "If you're bored, do this one."

PiotrKozlowski:
This is ready for testing but it was merged into #1963

uskka:
Works fine. Tested on katalyst01.
https://drive.google.com/file/d/1N06YF5VYjiKa4SORNcvnB9OMwZKRNnDh/view?usp=sharing

PR #1961, opened 2018-09-26
issue 1757 read solution after open

PiotrKozlowski:
#1757 

To force screen readers to read a text after hiting `Show Solution` we need to have an empty element with `aria-live="polite / assertive"` and content of this element shoud get updated after click.

- To check if this is working go to http://localhost:8000/contents/mwjClAV_@7.1:0KhpF2RH@21/Real-Numbers-Algebra-Essentials
- Start screen reader (Ubuntu - Windows + Alt + S)
- Click before `Show solution` and hit Tab to navigate to button - hit eneter - wait for screen reader to read text

### DIFF
--- a/src/scripts/modules/media/body/body.coffee
+++ b/src/scripts/modules/media/body/body.coffee
@@ -161,12 +161,11 @@ define (require) ->
 
           # Wrap solutions in a div so "Show/Hide Solutions" work
           $temp.find('.exercise .solution, [data-type="exercise"] [data-type="solution"]')
-          .wrapInner('<section class="ui-body">')
+          .wrapInner('<section class="ui-body" role="alert">')
           .prepend('''
             <div class="ui-toggle-wrapper">
               <button class="btn-link ui-toggle" title="Show/Hide Solution"></button>
             </div>''')
-          .append('<div class="ui-body-live" aria-live="polite"></div>')
 
           $temp.find('figure:has(> figcaption)').addClass('ui-has-child-figcaption')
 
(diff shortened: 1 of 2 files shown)


src/scripts/modules/media/body/body.coffee
```diff
@@ -460,10 +461,14 @@ define (require) ->
     toggleSolution: (e) ->
       $solution = $(e.currentTarget).closest('.solution, [data-type="solution"]')
       $solution.toggleClass('ui-solution-visible')
+      $uiBody = $solution[0].getElementsByClassName('ui-body')[0]
+      $uiBodyLive = $solution[0].getElementsByClassName('ui-body-live')[0]
       if $solution.hasClass('ui-solution-visible')
         $solution.attr('aria-expanded',true)
         $solution.attr('aria-label',"hide solution")
+        $uiBodyLive.innerHTML = $uiBody.innerHTML
```
philschatz: From https://stackoverflow.com/q/10349987 : Rather than creating a new element and moving the HTML, could adding `role=alert` to the element and then unhiding it have the same effect?

From:

> you can add on the fly or have it already on the hidden content, the change from display:none to display:block or the addition of content to the element with role=alert will cause an accessibility event to fire and the content to be announced. an example: [link] dl.dropbox.com/u/377471/tests/ajax.html – Steve Faulkner

Adrianltd: From: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role

>Because of its intrusive nature, the alert role must be used sparingly and only in situations >where the user's immediate attention is required. Dynamic changes that are less urgent should >use a less aggressive method, such as aria-live="polite" or other live region roles.

So I think that `aria-live="polite/assertive"` and setting `display: none;` instead of moving elements would be ok.

Issue #1971, opened 2018-10-07
Hide "Show More" Option When No Book Description is Present on the CNX Homepage


openstaxalina:
### Problem Description

**Before senna.syrena:**
![no show more](https://user-images.githubusercontent.com/7863733/46585368-ada2b880-ca35-11e8-8df6-b683897c127f.png)

**After senna.syrena, "Show More" appears when it shouldn't:**
<img width="1343" alt="screen shot 2018-10-07 at 1 34 31 pm" src="https://user-images.githubusercontent.com/7863733/46585376-be532e80-ca35-11e8-9209-7797b0f8ab02.png">

### Acceptance Criteria
-The "Show More" button does not appear if there is no description
-The "Show More" button does not appear if the description is very short, and fits within the default space.

openstaxalina:
@helenemccarron FYI this is the other issue we discussed.

Issue #1970, opened 2018-10-07
Re-implement ellipsis or other solution to show book descriptions on home page continue


openstaxalina:
### Problem description
With senna.syrena code, we lost the ellipses in the book descriptions on the home page:
**Before senna.syrena:**
![with ellipsis](https://user-images.githubusercontent.com/7863733/46585310-f1e18900-ca34-11e8-8e16-7f1fb7a2c885.png)

**After senna.syrena**
<img width="1272" alt="screen shot 2018-10-07 at 1 29 32 pm" src="https://user-images.githubusercontent.com/7863733/46585327-0b82d080-ca35-11e8-928a-a01926efabc3.png">

I believe this is because shave.js was removed/disabled.

###Acceptance Criteria
-Descriptions are cut off with an ellipsis or other way to show that they've been cut off
-shave.js has been completely removed from the code if this package will not be used.

openstaxalina:
@helenemccarron FYI, this is one of the issues we discussed.  The other one will be https://github.com/Connexions/webview/issues/1970

PiotrKozlowski:
After some research I've ended up with using almost pure css to show ellipsis. I've had to set `text-align: justify` to the descriptions. Let me know if you are okay with this solution. 

![shave](https://user-images.githubusercontent.com/31478212/46612665-815c6a00-cb11-11e8-9eac-8dc8159abc96.png)

openstaxalina:
@PiotrKozlowski I like this solution; thank you!  Could you please assign @Dantemss as a reviewer on your PR?  He wrote the original code that populates the book descriptions on the home page.

PR #1972, opened 2018-10-08
issues 1970 and 1971 - add ellipsis and remove .more button when desc is short enough

PiotrKozlowski:
#1970 #1971 

- remove some shave.js dependencies which are no longer in use
- add ellipsis (...) at the end of the line with some gradient effect
- hide `show more` button when description is short enough

Info: We couldn't use shave.js because this script was removing all html tags and we needed `<a>` tags from description.

shave

### DIFF
--- a/src/scripts/modules/featured-books/featured-books.coffee
+++ b/src/scripts/modules/featured-books/featured-books.coffee
@@ -1,5 +1,4 @@
 define (require) ->
-  _ = require('underscore')
   $ = require('jquery')
   require('jqueryui')
   Backbone = require('backbone')
@@ -19,13 +18,11 @@ define (require) ->
       super()
       @listenTo(featuredOpenStaxBooks, 'reset', @render)
       @listenTo(featuredCNXBooks, 'reset', @render)
-      @debouncedShaveBookDescriptionsAfterImagesLoaded = \
-        _.debounce(@shaveBookDescriptionsAfterImagesLoaded, 300)
 
     readMore: (event) ->
       read_more = $(this).parent()
       book = read_more.parent()
-      book.find('.description').css('height', 'auto')
+      book.find('.description').addClass('extended')
       $(this).hide()
       read_more.find('.less').show()
       event.preventDefault()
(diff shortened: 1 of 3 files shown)


src/scripts/modules/featured-books/featured-books.less
```diff
@@ -69,7 +69,33 @@
...
+            &:after {
+              display: none;
+            }
+          }
+        }
+
+        &.show-ellipsis {
+          position: relative;
+          &:after {
+            content: '...';
```
Dantemss: The previous plugin used `…` instead of `...`
Not sure it matters, but we'll have to adjust the automated tests to look for `...` instead if we leave this.

src/scripts/modules/featured-books/featured-books.coffee
```diff
@@ -45,7 +42,10 @@ define (require) ->
     onAfterRender: () ->
       $('.show-more-less .more').on('click', @readMore)
       $('.show-more-less .less').on('click', @readLess)
-      $(window).resize(@debouncedShaveBookDescriptionsAfterImagesLoaded)
-
-    onBeforeClose: () ->
-      $(window).off('resize', @debouncedShaveBookDescriptionsAfterImagesLoaded)
+      $('.description').each (i, desc) ->
+        $(desc).css('max-height', '100%')
+        if $(desc).height() <= 60
```
Dantemss: Won't this be a problem if it is checked before the images are loaded?
They tend to compress the text...
Might be fine, but have to test while simulating a slow connection with the dev console.

src/scripts/modules/featured-books/featured-books.less
PiotrKozlowski: Changed, thanks

src/scripts/modules/featured-books/featured-books.coffee
PiotrKozlowski: I've tested this with `Regular 3G` connection and it looks fine.
btw. When I was trying to simulate `Good 2G` nothing was loaded at all. Some scripts are timeouting.

Issue #1944, opened 2018-09-14
accessibility: Table of Contents should be links, not buttons


openstaxalina:
### Problem
For some reason, the software for the screen reader is not expanding Chapters to their sections in the navigation pane when you press enter. We understand it is because they're programmed as buttons instead of links. 
...
Navigation is difficult, as a lot of the table of contents navigation panel is coded using expandable “buttons” or “span” (not links). Screen readers won’t display these links correctly when they bring up their shortcut links menu.

### More Detail
From Phil: 

I think there are 2 problems:

1. the twisty is CSS and on a < span > which does not show up for screenreaders (it should be a link or button or something rather than a CSS ::before
2. the big blue [Contents + ] button should have aria labels that indicate something like "Click to Show the Table of Contents"

openstaxalina:
![image 2](https://user-images.githubusercontent.com/7863733/45575495-58f68e00-b839-11e8-81bb-20df5fbc83a0.png)
![image 3](https://user-images.githubusercontent.com/7863733/45575497-5ac05180-b839-11e8-90f1-250239db8383.png)

PiotrKozlowski:
Ready for testing in #1967

uskka:
@openstaxalina Works on default text to speech app on Ubuntu, doesn't work with Chromevox on Mac and Windows, also doesn't work with VoiceOver on Mac.

openstaxalina:
@uskka and @PiotrKozlowski thank you!

@PiotrKozlowski, do you know of any strategies that would make this functional on a broader set of screen readers?  I will also ask our accessibility group about this on Thursday.

PiotrKozlowski:
@openstaxalina I will try to fix this today.

uskka:
In Chromevox it reads only one random chapter, in voiceover on Mac it doesn't work at all.

PiotrKozlowski:
Hmm, for me it's working in incognito mode. My guess is Chromevox is not reading automatically links in lists which was already readed.

PiotrKozlowski:
Ready for testing at master

uskka:
@openstaxalina Looks good, but it reads only the first section when the list with section titles is opened. I attached a recording from the test. Could you please confirm that this is a desired behaviour for this issue? Recorded with Chromevox. 
https://drive.google.com/open?id=1UvjlRTSSxNPYMXVIy47FAma0NpdZHDH6

PiotrKozlowski:
[...]

PR #1975, opened 2018-10-10
Issue 1944v2

PiotrKozlowski:
#1944 

On Introduction page there are 2 `.toc`. One in sidebar and second at the bottom of page. Both are created from the same component so change in one was visible in both places.

When user click `Contents` attribute `role="alert"` will be added only to the `.toc` inside sidebar (`tocInsideSidebar = $('.toc')[0]`). This will force screen readers (chromevox) to read the toc.

Setting dynamically `role="alert"` to `<ul>` with modules after click on `section name` will also force screen readers to read this list.

This solution seems to work with ChromVox but not with default Ubuntu screen reader. Previous code was working only with Ubuntu. I've tried to combine both methods but then ChromVox had some trobules, and from what I know we have priority for supporting ChromeVox.

### DIFF
--- a/src/scripts/modules/media/nav/nav.coffee
+++ b/src/scripts/modules/media/nav/nav.coffee
@@ -67,6 +67,9 @@ define (require) ->
             @closeAllContainers(node.get('contents').models)
 
     toggleContents: (e) ->
+      // This component is used in the ToC as well as the intro page. 
+      // Having the aria attributes causes confusion because there are 2 `<nav>` elements on the page
+      // See #1975
       tocInsideSidebar = $('.toc')[0]
       if tocInsideSidebar.getAttribute('role')
         tocInsideSidebar.removeAttribute('role')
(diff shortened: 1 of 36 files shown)


brittweinstein:
CC @philschatz and @helenemccarron for review

src/scripts/modules/media/nav/nav.coffee
```diff
@@ -67,6 +67,11 @@ define (require) ->
             @closeAllContainers(node.get('contents').models)
 
     toggleContents: (e) ->
+      tocInsideSidebar = $('.toc')[0]
```
philschatz: Could this flag be passed in as an argument into the constructor? That would help clarify that this component is used in multiple places with different settings.

If that requires too much work (like passing an extra variable through many objects) then at least a comment stating why this extra code is needed. Example:

```suggestion
      // This component is used in the ToC as well as the intro page. 
      // Having the aria attributes causes confusion because there are 2 `<nav>` elements on the page
      // See #1975
      tocInsideSidebar = $('.toc')[0]
```

PiotrKozlowski:
@philschatz Thanks, I've applied requested changes

Issue #1896, opened 2018-09-04
hyperlinked TOC: links should go to top of module


kerwinso:
Related to #1838. 

We noticed that clicking on a link in the new TOC jumps you to where the content starts, but below the header — so the user can’t tell which module number they landed on. The preferred experience is to land at the top of the module, or at least at a location where the module number and title are visible.

![image](https://user-images.githubusercontent.com/11298165/45050234-52e8fc00-b046-11e8-8c33-8ed661100eb1.png)

openstaxalina:
Update:
Issue occurs in *Chrome*, not in Safari or Firefox.
Issue does not occur from left-hand ToC navigation pane.
Issue occurs only when first navigating to a module, either from the hyperlinked ToC in the intro page body, or clicking "Next" or "Back".

openstaxalina:
Here are Tom's notes on this issue:
https://openstax.slack.com/archives/C0LA54Q5C/p1536174056000100

PiotrKozlowski:
It works correctly after setting `height: 160px` to `<header id="header">`. I'm not sure why, because `<header>` doesn't seem to rerender after page change and it's height seems to be `160px` all the time anyway. I tried to set `min-height: 160px` but it wasn't working. It has to be `height: 160px`.

uskka:
Works fine. Tested on katalyst01 using Chrome.
https://drive.google.com/file/d/18ymNs5kYkzFQT-TxLonB54YVWjl98XnY/view?usp=sharing

PR #1982, opened 2018-10-16
issue 1896 disable overflow-anchor (Google Chrome anchor scrolling)

PiotrKozlowski:
#1896 

Issue was caused because Google Chrome Anchor Scrolling feature.

At every page change content of `<header id="header">` is rerendered which mean height of this element i jumping from `0` to `x`. When we change page for the first time it tooks longer time to load than next changes which mean view is scrolled before height of header was adjusted, so Google Chrome is trying to fix this with new feature called Anchor Scrolling. 

Fix is simple - disable anchor scrolling. However i'm not sure were should I put this css?

### DIFF
--- a/src/scripts/pages/app/app.less
+++ b/src/scripts/pages/app/app.less
@@ -1,5 +1,9 @@
 @import "../../../styles/variables.less";
 
+body {
+  overflow-anchor: none;
+}
+
 #header {
   max-width: @max-viewport-width;
   background-color: white;


src/scripts/pages/app/app.less
```diff
@@ -1,5 +1,9 @@
 @import "../../../styles/variables.less";
 
+body {
+  overflow-anchor: none;
```
philschatz: ooh, I didn't know about [overflow-anchor](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-anchor). Nice!

helenemccarron: Same here 😁

Issue #2000, opened 2018-10-23
mobile device: some buttons/text overlap in portrait orient


omehes:
Steps:

1. run staging.cnx.org on a smaller mobile device (I used Android phone LG G3)
2. open a book and scroll to the bottom of the page
3. in portrait orient, tap on "More information" button
Expected: "More information" section opens and buttons/text remain displayed organised and side-by-side
Actual: Some buttons and text overlap in portrait orient (see attached screen shot). This applies to any book on smaller mob devices.

I tried on Android phone LG G3 on Chrome

Link: https://staging.cnx.org/contents/AndhM9Sd@4.13:HCaG_W5c@4/Introduction

Screen shot: 
![Screenshot_2018-10-23-12-28-31.png](https://images.zenhubusercontent.com/5bc8d1f2fe40ab78d94e65fb/9594f9ae-5369-49f7-b8bb-73a7c37c714e)

Note that something similar happens with Attribution and History buttons:

![Screenshot_2018-10-23-12-45-51.png](https://images.zenhubusercontent.com/5bc8d1f2fe40ab78d94e65fb/d3b1318b-4aec-4ba5-9a3a-c72d899811c0)

uskka:
In History section issue was fixed but still doesn't work for Attribution.

![44753988_282336835725785_3779686432977715200_n](https://user-images.githubusercontent.com/40228918/47517084-6155d600-d887-11e8-9a7f-f15d3bc11758.jpg)
![44811200_2160619457336155_7979437924682825728_n](https://user-images.githubusercontent.com/40228918/47517086-61ee6c80-d887-11e8-8fa3-d89c956311e8.jpg)

uskka:
Looks good. Tested on katalyst01.
![screenshot_20181026-121521_firefox](https://user-images.githubusercontent.com/40228918/47560614-0b317300-d919-11e8-9dbe-5d918ffe5427.jpg)

omehes:
I am re-opening this issue as I verified it on qa.cnx.org and staging.cnx.org. Still not fixed (or merged). Waiting for the fix.
@edwoodward @brittweinstein @uskka

PiotrKozlowski:
@omehes It' working on http://katalyst01.cnx.org/
I can't see mine code on qa.cnx.org and staging.cnx.org which mean deploy wasn't succesfull or something else happend.

omehes:
[...]

PR #2002, opened 2018-10-24
fix responsivity for metadata

PiotrKozlowski:
#2000 

Aftert:

### DIFF
--- a/src/scripts/modules/media/footer/footer.less
+++ b/src/scripts/modules/media/footer/footer.less
@@ -21,6 +21,7 @@
       display: inline-block;
       padding: 0.7rem 1rem;
       margin-right: 0.3rem;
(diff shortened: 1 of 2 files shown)


src/scripts/modules/media/footer/footer.less
```diff
@@ -21,6 +21,7 @@
       display: inline-block;
       padding: 0.7rem 1rem;
       margin-right: 0.3rem;
+      margin-top: 3px;
```
Adrianltd: Can we use `rem` here, since margin-right and paddings are in rems too?

helenemccarron: I agree, that would be a good idea to keep the unit consistent

PR #2007, opened 2018-10-25
issue 1944v3 rewrite accessibility support for links inside toc 

PiotrKozlowski:
#1944 

Previous version worked only in incognito mode and weird things were happening in normal mode.

**Changes:**
Remove `aria-live` and `role="alert"` attributes from `.toc` and set `focus()` for closest link or button instead.

In `.toc` there are: `button.section-wrapper` for collapsing sections of modules and `.name-wrapper a` for individual module. I'm selecting first of those two with `querySelector` which is forcing ChromeVox to read exactly this text.

### DIFF
--- a/src/scripts/modules/media/tabs/contents/toc/section.coffee
+++ b/src/scripts/modules/media/tabs/contents/toc/section.coffee
@@ -83,10 +83,12 @@ define (require) ->
         e.preventDefault()
         @toggleSection()
         isExpanded = @$el.find('button')[0].getAttribute('aria-expanded')
-        if isExpanded == 'true'
-          @$el.find('ul')[0].querySelector('.name-wrapper a, .section-wrapper').focus()
-        else if isExpanded == 'false'
-          @$el.find('> div > .section-wrapper')[0].focus()
+        if isExpanded is 'true'
+          focusEl = @$el.find('ul')[0].querySelector('.name-wrapper a, .section-wrapper')
+        else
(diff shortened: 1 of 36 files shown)


src/scripts/modules/media/tabs/contents/toc/section.coffee
```diff
@@ -82,8 +82,11 @@ define (require) ->
       if e.keyCode is 13 or e.keyCode is 32
         e.preventDefault()
         @toggleSection()
-        @$el.find('> div > .section-wrapper').focus()
-        @$el.find('> ul').attr('role', 'alert')
+        isExpanded = @$el.find('button')[0].getAttribute('aria-expanded')
+        if isExpanded == 'true'
+          @$el.find('ul')[0].querySelector('.name-wrapper a, .section-wrapper').focus()
+        else if isExpanded == 'false'
```
philschatz: Is there an `else` case or does this code assume that the `aria-expanded` attribute is _always_ present on the button? (I jut worry about accidental code fall-through) 

Maybe something like:

```coffeescript
if isExpanded is 'true'
    focusEl = @$el.find('ul')[0].querySelector('.name-wrapper a, .section-wrapper')
else
    focusEl = @$el.find('> div > .section-wrapper')[0]

focusEl.focus()
```

PiotrKozlowski: @philschatz Actually this function is called only on element which have `aria-attribute` but your version looks better :D

Issue #1999, opened 2018-10-23
Mobile devices: Some text can go off screen in both orients


omehes:
Steps:
1. run staging.cnx.org on a smaller mobile device (I tested on Android phone LG G3) 
2. open a book, like https://staging.cnx.org/contents/AndhM9Sd@4.13:QQLN142k@8/Dividing-Polynomials
3. scroll around the page in both orients
Expected: the text should be readable in both orients and should not be cut-off
Actual: Some text, like math equations or tables can have end parts cut-off in portrait or both orients. See examples below:

1. Link: https://staging.cnx.org/contents/AndhM9Sd@4.13:QQLN142k@8/Dividing-Polynomials
- in this case the table is so wide that its right section (part of Function column and all of f(c) column) is hidden in both orient, making it unreadable for the user.

![Screenshot_2018-10-22-14-52-11.png](https://images.zenhubusercontent.com/5bc8d1f2fe40ab78d94e65fb/c3f5e1da-cb0e-4f5b-8811-680062508b75)

2. Link: https://staging.cnx.org/contents/AndhM9Sd@4.13:2oR4tJO8@6/Polynomial-Equations
- text is cut-off in portrait orient only

![Screenshot_2018-10-22-14-37-42.png](https://images.zenhubusercontent.com/5bc8d1f2fe40ab78d94e65fb/71c8df5d-7873-4081-ab0c-cdf0998b7eac)

3. Another example: https://staging.cnx.org/contents/AndhM9Sd@4.13:gc_Ek8hX@2/Sequences
- solution section is shifted off the screen in portrait orient

![Screenshot_2018-10-22-15-51-06.png](https://images.zenhubusercontent.com/5bc8d1f2fe40ab78d94e65fb/5c1a8faa-8fe6-4148-b185-a81a486ba6f4)

@brittweinstein

edwoodward:
@PiotrKozlowski We were wondering if it is possible to get MathJax to wrap the math that is an issue. Please give it a try.

omehes:
Hi,

on mobile devices I cannot scroll the tables, like in  
https://staging.cnx.org/contents/AndhM9Sd@4.13:QQLN142k@8/Dividing-Polynomials

If student uses a phone to read the book, some sections will be lost  
in both orients (in some cases only in Portrait).

Ottó



Quoting Piotr Kozłowski <notifications@github.com>:

PiotrKozlowski:
[...]

PR #2010, opened 2018-10-26
Issue 1999

PiotrKozlowski:
#1999 

`overflow: auto;` for `[data-type="equation"]` which makes blockish equations scrollable on smaller screens.

Wrap tables inside unbaked books with additional div so we can apply `overflow-x: auto` to them which makes them scrollable for devices with smaller screens.


philschatz:
It seems like there are extra commits included in this PR. Should this be a PR to `master` or to another branch?

PiotrKozlowski:
@philschatz This was targeted to `katAW.u..u..` but it's closed already.
Should I create new branch and pr with only those two commits `overflow:auto for data-type=equation`, `make tables inside unbaked books scrollable`?

Issue #2013, opened 2018-10-26
Read page content after page change


PiotrKozlowski:
After changing page inside book there should be some kind of information for screen readers indicating content is loading and when it's done we should set `focus()` on page title.

After changing page with `enter` inside TOC or with `next` / `prev` buttons, focus should be at progress bar with message "Loading" and when content is loaded it should read the title.

<img width="1277" alt="screen shot 2018-10-26 at 12 12 02" src="https://user-images.githubusercontent.com/31478212/47572266-ddf5bc80-d93a-11e8-8184-8ac72bcc1c96.png">

![screenshot-katalyst01 cnx org-2018 10 26-12-12-50](https://user-images.githubusercontent.com/31478212/47572471-578daa80-d93b-11e8-8349-3ef8c885441e.png)

uskka:
Works fine. Tested with Chromevox. 
https://drive.google.com/file/d/13iYS82RP8i98m4csTXUYV8jMrADXTQpw/view?usp=sharing

PR #2043, opened 2018-11-06
issue 2013 read load message and page title when user change page with keyboard

PiotrKozlowski:
#2013 

This is accessibility feature.

**Assumption:** When user navigate trough `Table of Contents` and hit `enter` then screen reader should read `Loading` message and after content is loaded user should be informed about this and focus should change to page title. The same should happens after selecting `Back` and `Next` page.

### DIFF
--- a/.jenkins/deploy
+++ b/.jenkins/deploy
@@ -1,3 +1,5 @@
+@Library('pipeline-library') _
+
 pipeline {
   agent { label 'docker' }
   environment {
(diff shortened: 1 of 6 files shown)


PiotrKozlowski:
It seems like `onRender()` function inside `src/scripts/modules/media/header/header.coffee` is triggered multiple times and twice when `@model.asPage()?.get('loaded')` so I can't set `window.pageWasChangedWithKeyboard = false` because when I do this then title is rerendered and focus will be never set.
Now this works correctly but if user will change page with keyboard then after every page change (with mouse click) focus will be set to page title.

Issue #2042, opened 2018-11-05
Accounting - Alignment within table cells in webview


mwest-WWprod:
content03
GLOBAL - Webview

We have applied tagging to content in table cells in order to right-align numerals, however the output is not appearing as expected. This alignment is particularly important due to the nature of accounting. Same issue as seen here for PDF output [#3805](https://github.com/Connexions/oer.exports/issues/3805).

![webviewtable_alignment](https://user-images.githubusercontent.com/19212926/48024898-de970b80-e10f-11e8-86ee-d666a60eed6f.JPG)

Example given above can be found [here](https://content03.cnx.org/contents/QvVVZeAI@7.1:G7ly_b1W@7.1/Problem-Set-A).

brittany-johnson:
![screen shot 2018-11-06 at 10 43 05 am](https://user-images.githubusercontent.com/24909007/48079339-c88d5780-e1b0-11e8-9121-0a29d5471391.png)

Stackblocks:
Displaying as specified with updated version of content / code.

![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/84851dd0-46c7-48c0-8186-af4e05b53b12)

PR #2044, opened 2018-11-06
issue-2042

brittany-johnson:
https://github.com/Connexions/webview/issues/2042

### DIFF
--- a/src/scripts/modules/media/body/content-common.less
+++ b/src/scripts/modules/media/body/content-common.less
@@ -377,7 +377,11 @@ table {
 
   // Cells
   thead,
-  tbody,
+  tbody {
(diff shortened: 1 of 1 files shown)


src/scripts/modules/media/body/content-common.less
```diff
@@ -377,7 +377,11 @@ table {
 
   // Cells
   thead,
-  tbody,
+  tbody {
```
helenemccarron: Should this apply to all books or only to baked books?

brittany-johnson: I would say all books. If it has the `data-align="right"` I just want to make sure that it's actually aligning right everywhere @helenemccarron

Issue #2038, opened 2018-11-05
Numbering by Step #


wilkus27:
In _Elementary Algebra_ in **How To** note and in section **Key Concepts** there are lists which should be numbered by _Step 1, Step 2 etc._

webview (e.g. https://katalyst01.cnx.org/contents/CImQfPDv@3.14:Tj87ETCR@4/3-1-Use-a-Problem-Solving-Strategy)

![step_how_to](https://user-images.githubusercontent.com/40228854/48010120-4738b500-e11d-11e8-8b36-dcdb39309c6e.JPG)

PDF:
![step_how_to_pdf](https://user-images.githubusercontent.com/40228854/48010170-5d467580-e11d-11e8-8792-a27b918dd1fe.JPG)

webview (e.g. https://katalyst01.cnx.org/contents/CImQfPDv@3.14:7BOr1QuT@3.14/Key-Concepts)

![step_key](https://user-images.githubusercontent.com/40228854/48010210-72230900-e11d-11e8-80f8-aad07db06021.JPG)

PDF:
![step_key_pdf](https://user-images.githubusercontent.com/40228854/48010235-7d763480-e11d-11e8-82d6-e678d41a6a7d.JPG)

openstaxalina:
Hopefully these have the `stepwise` class on them; we should check in sprint planning.

lwtchu:
@openstaxalina they do :)!
![screen shot 2018-11-07 at 3 57 22 pm](https://user-images.githubusercontent.com/11299125/48163637-da045b80-e2a5-11e8-9edf-c50638ec1e16.png)

openstaxalina:
Thanks @lwtchu !
In that case, @PatBoj could you use that class to add the word "Step" before each number?  I think this is what we decided to do with all lists with that class.

cc @helenemccarron

openstaxalina:
Issue moved to [Connexions/cnx-recipes #704](https://github.com/Connexions/cnx-recipes/issues/704) via [**ZenHub**](https://www.zenhub.com/)

Issue #2031, opened 2018-11-01
WAITING WW REVIEW: Accounting - Leading in webview lists


brittany-johnson:
@mwest-WWprod commented on [Mon Oct 15 2018](https://github.com/Connexions/cnx-recipes/issues/679)

content03
GLOBAL

Lists in webview have inconsistent leading between items.

![listspacing_webview](https://user-images.githubusercontent.com/19212926/46982618-f77d4580-d0aa-11e8-9fb6-116bdf5a5cf6.png)

---

@openstaxalina commented on [Tue Oct 16 2018](https://github.com/Connexions/cnx-recipes/issues/679#issuecomment-430301477)

@oscryan would you mind reviewing the content in this case?  I have never seen this kind of styling issue with a list.

If content is correct, please post a screenshot of the markup with a link, and remove your assignment.

cc @lwtchu 

---

@oscryan commented on [Tue Oct 16 2018](https://github.com/Connexions/cnx-recipes/issues/679#issuecomment-430309259)

The content is treated differently in each list item. The first item has the text inside only the item tag; the second item has two paragraph tags nested in the list tag. The third has no nesting, like the first. This seems to have caused the item number to center vertically in the third case, but I don't know why it's different from the first.

---

@oscryan commented on [Tue Oct 16 2018](https://github.com/Connexions/cnx-recipes/issues/679#issuecomment-430309807)

https://legacy-content03.cnx.org/content/m10005/latest
<img width="1115" alt="screen shot 2018-10-16 at 11 41 50 am" src="https://user-images.githubusercontent.com/1895432/47032674-8edcaa00-d138-11e8-8f76-093d2d14efa0.png">

<blockquote><img src="https://legacy-content03.cnx.org/book_icon_cnx.png" width="48" align="right"><div>Portal</div><div><strong><a href="https://legacy-content03.cnx.org/content/m10005/latest/">Compute a Predetermined Overhead Rate and Apply Overhead to Production</a></strong></div><div>Job order cost systems maintain the actual direct materials and direct labor for each individual job. Since production consists of overhead—indirect materials, indirect labor, and other overhead—we need a methodology for applying that overhead. Unfortunately, the nature of indirect material, indirect labor, and other overhead expenses makes it impossible to determine the exact amount of overhead for each specific job. For example, how do you know the cost of electricity and heat for manufacturing one job? And, if you did, is it fair to say products manufactured in January are more expensive than the same product manufactured in March because of heat expense?</div></blockquote>

---

[...]

PR #2046, opened 2018-11-08
Add "Step" before list with "stepwise" calss

PatBoj:
Issue #2038 

## **Before:**
image

## **After:**
image

Those large margins between list elements are caused by this PR https://github.com/Connexions/webview/pull/2041. I mentioned about this in this issue #2031.

### DIFF
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -31,6 +31,16 @@ ul, ol {
   padding-left: 1rem;
 }
 
+.stepwise {
+  list-style-type: none;
+  counter-reset: stepwiseCounter;
+  padding-left: 0;
+  > li:before {
+      content: "Step " counter(stepwiseCounter) ". ";
+      counter-increment: stepwiseCounter;
+  }
+}
+
 [data-bullet-style="none"] {
   list-style-type: none;
 }


src/scripts/modules/media/body/content-baked.less
```diff
@@ -31,6 +31,16 @@ ul, ol {
...
 
+.stepwise {
+  list-style-type: none;
+  counter-reset: stepwiseCounter;
+  padding-left: 0;
+  > li:before {
+      content: "Step " counter(stepwiseCounter) ". ";
+      counter-increment: stepwiseCounter;
+  }
+}
```
PiotrKozlowski: I think you can delete margin for list items if you think it shouldn't be here:
```suggestion
 > li {
       margin-bottom: 0;
   }
}
```

PatBoj: Yes, what about list without `class="stepwise"` like here?
image

PiotrKozlowski: I see some options:
1. Remove global margin for `<li>` and set margin depends of parent's class.
2. Leave global margin as it is and set custom margin for `<li>` depends of parent's class.
3. Lower global margin, ex: from `1em` to `0.8em` so change is not visible that much.

Issue #2006, opened 2018-10-24
mobile devices: primary text with too large font size overlaps with the image 


omehes:
Steps:

1. run staging.cnx.org on smaller device (LG G3, Samsung S5 phones, iPhone X, etc)
2. check the main page (the image and the title text)
Expected: Image and primary title text should be organised similar way as on desktops (text to the left of the main image subject)
Actual: Primary title text overlaps with the image and has incorrect height (see screen shots)

Screen shot on iPhone X:
![Screen Shot 2018-10-24 at 1.16.01 PM.png](https://images.zenhubusercontent.com/5bc8d1f2fe40ab78d94e65fb/5601b446-5b24-4945-8285-446cbea59706)

Screen shot on Desktop (correct):
![Screen Shot 2018-10-24 at 2.28.43 PM.png](https://images.zenhubusercontent.com/5bc8d1f2fe40ab78d94e65fb/7622be9c-8430-4eda-becb-eb1a71e14476)

edwoodward:
@omehes I think the secondary line of text is hidden on purpose on mobile. The height of the primary text on the image is definitely an issue.

PiotrKozlowski:
It's very hard to do something with this text because of background image, but I've added white overlay above it and some white shadow around text to make it more readable, I've also aligned text to the left but I didn't change font-size, because smaller size is even harder to read at this background.

Let me know if this is okay.

![homeresp](https://user-images.githubusercontent.com/31478212/47649969-1e457c80-db7f-11e8-9f27-4a717e29653c.png)

omehes:
@PiotrKozlowski it is better now but that empty space above the text... It feels like everything is shifted to the bottom. From what you are saying, it is not easy to shift the text upwards and use that empty space for the text?

PiotrKozlowski:
@omehes I've moved this text to the bottom so it not overlap with her face, but I can actually move it to the top. Is this better:?

![homeresp](https://user-images.githubusercontent.com/31478212/47659286-eba67e80-db94-11e8-82b3-5ec878ea7370.png)

omehes:
@PiotrKozlowski definitely better.  I will need to test it on various device screen sizes. I guess, I could push the fix to qa and test it. Not sure how all this works yet, I am still learning the processes :)

PiotrKozlowski:
@omehes I've merged changes to `master` so you should be able to deploy and check :)

openstaxalina:
[...]

PR #2047, opened 2018-11-08
issue 2006 fix resposivity issue between firefox and chrome for splash text

PiotrKozlowski:
#2006 

It seems like Firefox and Google Chrome were calculating differently position and height of `relative` elements so I've rewrote this part. There are screenshots in issue conversation and you can also check previous version at `qa.cnx.org`.


PiotrKozlowski:
@Adrianltd I think we should do this, however that might be tricky because we will have to check style of each element to be sure that nothing was broken.

Issue #2048, opened 2018-11-08
"section section" is read before the section title


uskka:
When any text to speech app is used, and we open a new section, before section title is read, we can hear "section section". 
Here is a reproduction of this isse with Chromevox: 
https://drive.google.com/file/d/13iYS82RP8i98m4csTXUYV8jMrADXTQpw/view

PiotrKozlowski:
It's caused because `header.coffee` is rerendered after it's loaded.
It looks like this:
1. header is loaded:
2. focus title
3. read `section - ...`
4. header is rerendered after first word was red
5. focus is lost
6. header is loaded again
7. focus is set to title
8. read `section - title`

I'm not sure why `header.coffee` is rendered multiple times. Also this happens only on first load on page. If you will select this page again later it will be loaded normally and `header` will be loaded only once.

Stackblocks:
Unable to test because I don't have access to my accessibility machine today. Please test this. Thank you.

PR #2070, opened 2018-11-16
issue 2048 do not rerender header after it's loaded

PiotrKozlowski:
#2048 

- Do not trigger `@render` when header is already loaded.
- Reset `pageWasChangedWithKeyboard` variable after `focustTitle()` was called.

However I'm not sure why in the first place `@render` was called at `change:currentPage.loaded` - it was causing additional render of header which was why ChromeVox was reading `section` twice.

### DIFF
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -26,6 +26,11 @@ ul, ol {
   margin-top: 1rem;
 }
 
+.stepwise {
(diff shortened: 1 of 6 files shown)


philschatz:
Nice sleuthing! I am not sure why `@render` was called at `change:currentPage.loaded`. It might have been added "just to be sure the page was rendered". If that is not found to be the case then there does not seem to be a need for it.

Issue #2055, opened 2018-11-12
"This page is not in any books" empty page is scrollable


omehes:
Steps:

1. run staging.cnx.org on (mobile) device
2. open a single content page (not a book), for example: https://staging.cnx.org/contents/QILbGRy4@3/Hearing-Harmony-What-is-Harmony
3. when you get the "This page is not in any books" screen, start scrolling that empty page
Actual: You can scroll the empty page
Expected: The empty page does not need to be scrollable

Checked on mobile Android LG G3 and other simulated devices and desktop browsers

@openstaxalina Opening this bug as per discussion with @PiotrKozlowski in bug https://github.com/Connexions/webview/issues/2021

reedstrm:
If there _is_ results in that pane, is it also too large? I.e. scrollable to extra empty space at the bottom?

PiotrKozlowski:
Merged into master

omehes:
Verified on qa.cnx.org on LG G3 and simulated mob devices. Fixed

{"date": "2018-11-20 12:14:23 CST",
 "webview": "a05781ee00cde05e5ce04d017c361d7a3cde1722 (v0.41.1-33-ga05781e)",
 "cnx-archive": "582474ce6a01415179dcbbb982bcc469fd870b58",
 "cnx-publishing": "6d86c1f9f797119d0bb9302a33ca1cfebba86005",
 "oer.exports": "vagrant.valkyrie v0.66.0",
 "press": "7.1.0",
 "cnx-deploy": "v2.31.0-21-g5c78537"
}

omehes:
verified in staging.cnx.org book https://staging.cnx.org/contents/OlIuW7sM@1/Derived-copy-of-The-Circle-of-Fifths

{"date": "2018-11-28 17:45:22 CST",
 "webview": "5512a7c1a8bcee7f5940fdbb5e75be44cb7e1583 (v0.41.1)",
 "cnx-archive": "4.6.0",
 "cnx-publishing": "0.14.1",
 "oer.exports": "vagrant.valkyrie v0.66.0",
 "press": "7.0.1",
 "cnx-deploy": "v2.31.2"
}

PR #2076, opened 2018-11-19
issue 2055 replace visibility:hidden with other css formulas so it's not alocati…

PiotrKozlowski:
…ng space on mobile

#2055 

`visibility: hidden;` is hiding content but it's also allocating space for this content which was why you could scroll empty page. Setting `display: none;` will cause "jumping content" effect so I've used other css trick.

### DIFF
--- a/src/scripts/modules/window-with-sidebar/window-with-sidebar.less
+++ b/src/scripts/modules/window-with-sidebar/window-with-sidebar.less
@@ -50,7 +50,11 @@
     > .main {
       width: ~"calc(100% -" @sidebar-width + @fudge ~")";
       @media (max-width: 640px) {
-        visibility:hidden;
+        position: absolute; 
+        overflow: hidden; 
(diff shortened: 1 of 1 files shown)


src/scripts/modules/window-with-sidebar/window-with-sidebar.less
```diff
@@ -50,7 +50,11 @@
     > .main {
       width: ~"calc(100% -" @sidebar-width + @fudge ~")";
       @media (max-width: 640px) {
-        visibility:hidden;
+        position: absolute;
```
helenemccarron: how about using `display: none` instead of `visibility: hidden`? That would remove it from the flow of the document.

PiotrKozlowski: Done. I've had to change `display: inline-block;` to `block` for `.main` and now it's not "jumping".

Issue #2093, opened 2018-11-27
Alignment of sentences in Step lists


wilkus27:
On katalyst01 (e.g. https://katalyst01.cnx.org/contents/AndhM9Sd@4.13:mbIpaplX@4/1-4-Decimals) multiline sentences in lists with Step item should be aligned to the first line, not to the word **Step**.

webview
![how_to](https://user-images.githubusercontent.com/40228854/49087265-d76aa580-f256-11e8-9783-aab134409221.JPG)

PDF:
![how_to_pdf](https://user-images.githubusercontent.com/40228854/49087285-de91b380-f256-11e8-8465-8aafb2095ef5.JPG)

wilkus27:
Since step lists was added to Chemistry (https://github.com/Connexions/cnx-recipes/pull/744) those changes impact also on this book.

PR #2095, opened 2018-11-27
issue 2093 fix for displaing list items for ul, ol with stepwise class

PiotrKozlowski:
#2093 

@PatBoj This change will require you to wrap everything except token in additional wrapper. Is this doable?
Now its like this:
```
ul.stepwise
  li
    .os-stepwise-token Step 2.
    Some text here
```
I would like it to look like this:
```
ul.stepwise
  li
    .os-stepwise-token Step 2.
    .os-stepwise-content Some text here
```

Because my solution will work until there is list like this:
```
ul.stepwise
  li
    .os-stepwise-token Step 2.
    Some text
    ul Another nested list
    Some text
```
In case like this flexbox will alocate space for 4 columns instead of two

Works for example from issue:
Before:
screenshot_2018-11-27 intermediate algebra - openstax cnx 1

After:
screenshot_2018-11-27 intermediate algebra - openstax cnx

Not working for another example because text need additional wrapper:
Before:
screenshot_2018-11-27 intermediate algebra - openstax cnx 3

After:
screenshot_2018-11-27 intermediate algebra - openstax cnx 2

With additional `.os-stepwise-content` wrapper (I should also remove padding left from nested lists):
screenshot_2018-11-27 intermediate algebra - openstax cnx 4

### DIFF
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -29,6 +29,12 @@ ul, ol {
 .stepwise {
   list-style-type: none;
   padding-left: 0;
+  > li {
+    display: flex;
+    .os-stepwise-token {
+      white-space: pre;
+    }
+  }
 }
 
 .circled {


PiotrKozlowski:
@PatBoj Awesome, thanks! :)

PR #2103, opened 2018-12-03
switch to use settings.json

philschatz:
rather than `settings.js`. This seems to work locally but I am not sure if anything was missed.

## Changes

- change the string to import `settings` in many files
- change grunt so it does not ignore `settings.js` anymore
- include the `requirejs-json` package
- change the tarball-creation-scripts to build a JSON file
- move the “supported languages” into `settings.json` (not sure why they were included in `settings.js`)

Also, `script/start` did not exist & `nginx.development.conf` no longer exists so I had to do the following to test locally:

1. install a simple static HTTP server
1. `cp bower_components/ src/bower_components/`
1. `http-server ./src/`

and then open up a browser. I only verified that there were no 404's on the main page.

Also, I updated `script/gen-settings-from-env` but I am not sure if I got everything.


brittweinstein:
Super low prio. Will reopen if needed.

Issue #2102, opened 2018-12-03
Remove bullet lists from Preface


wilkus27:
On katalyst01 (https://katalyst01.cnx.org/contents/Sr8Ev5Og@9.4:F_mjYFfh@20/Preface) dots from bullet lists should be removed:

webview:

![dots](https://user-images.githubusercontent.com/40228854/49372462-0cfe0b80-f6fb-11e8-8cb1-314b8f1b964b.JPG)

PDF:

![dots_pdf](https://user-images.githubusercontent.com/40228854/49372484-1f784500-f6fb-11e8-9434-c159cbef85ff.JPG)

Adrianltd:
The first list has in the content `data-bullet-style="bullet"`.
Another one can be fixed in code. Actually that fix will fix both, but `data-bullet-style="bullet"` will remain in the first list, which will be incorrect then and might cause issues in the future.

helenemccarron:
@openstaxalina I think it would be best to change the bullet style in content then make sure the css targets the `data-bullet-style` properly. Currently, @Adrianltd's fix will change all list to not have bullets in recipes. I think that might cause unintended consequences later.

openstaxalina:
Yep, let's avoid that broad of a change.  We can adjust this in content.

openstaxalina:
Issue moved to [Connexions/content-managers #500](https://github.com/Connexions/content-managers/issues/500) via [**ZenHub**](https://www.zenhub.com/)

PR #2109, opened 2018-12-05
removed bullet-list styling from lists in the preface

Adrianltd:
https://github.com/Connexions/webview/issues/2102


src/scripts/modules/media/body/content-baked.less
```diff
@@ -26,6 +26,12 @@ ul, ol {
   margin-top: 1rem;
 }
 
+.preface {
```
helenemccarron: see https://github.com/Connexions/webview/issues/2102#issuecomment-444571730

Issue #2097, opened 2018-11-29
Lists and figures out of the padding


brittweinstein:
@openstaxalina commented on [Thu Nov 08 2018](https://github.com/Connexions/content-managers/issues/464)

@wilkus27 commented on [Mon Nov 05 2018](https://github.com/Connexions/webview/issues/2040)

On katalyst01( e.g. https://katalyst01.cnx.org/contents/CImQfPDv@3.14:Z8sTwcjE@4/3-3-Solve-Mixture-Applications - Example 3.26) in Show Solution lists and figures are too far to the left:

![out](https://user-images.githubusercontent.com/40228854/48011567-54a36e80-e120-11e8-8e6d-36cf1727dfb9.JPG)

---

@lwtchu commented on [Wed Nov 07 2018](https://github.com/Connexions/webview/issues/2040#issuecomment-436790196)

@wilkus27 is this global or just once instance? 

---

@wilkus27 commented on [Thu Nov 08 2018](https://github.com/Connexions/webview/issues/2040#issuecomment-436911000)

@lwtchu It is a global issue.

---

@lwtchu commented on [Thu Nov 08 2018](https://github.com/Connexions/webview/issues/2040#issuecomment-437066958)

@wilkus27 would it be too difficult to list all instances? if it is, we can treat this as a known issue. it's just messy mark up that needs to be cleaned up i think.

---

@wilkus27 commented on [Thu Nov 08 2018](https://github.com/Connexions/webview/issues/2040#issuecomment-437074913)

@lwtchu It's possible to do, but there will be a lot of instances and it is very likely that this issue also occurs in two other Algebra books (Prealgebra, Intermediate Algebra) so it could be a very long list of all cases. But  in the other hand maybe we can use Stefan's tool, so it will be easier to find this instances. @uskka we can find **Examples** by your tool. That's correct?

---

@openstaxalina commented on [Thu Nov 08 2018](https://github.com/Connexions/webview/issues/2040#issuecomment-437172257)

I think this is not worth more time investigating right now, due to the time involved to locate and fix these instances.  I am going to move this to content-managers in the backlog.  Thank you very much @wilkus27 and @lwtchu for the investigation on this!

---

@wilkus27 commented on [Tue Nov 13 2018](https://github.com/Connexions/content-managers/issues/464#issuecomment-438281545)

[...]

PR #2111, opened 2018-12-05
removed paddings from p, ul, ol and .os-table in show solution area

Adrianltd:
https://github.com/Connexions/webview/issues/2097

After removing paddings of `p`, `ul`, `ol` and `.os-table` the solutions for Examples look fine to me.
I'm not sure if that should be removed only for `examples` or also as is now for `notes` and `exercises`?


Adrianltd:
image

Adrianltd:
image

PR #2120, opened 2018-12-07
need the | for continuation lines

reedstrm:
Seems the ftl format is very fragile, and if you break it, all the translations break.

I see a completely _missing_ "get this book" button with the broken FTL, as opposed to the broken styling of that button for devb. You can verify that this is a l20n.js problem noticing that 1. the menu entries at the top of the page are missing, and 2. change browser language to Polish, and everything is fixed. In the console, there's traceback from l20n.js 

With this fix, everything works


helenemccarron:
That's "fragile" indeed...

PR #2139, opened 2018-12-18
Style .os-stepwise

PiotrKozlowski:
@PatBoj pr https://github.com/Connexions/cnx-recipes/pull/750 was merged so now we can apply styles for `.os-stepwise`

This style was removed due to style breaking in books which wasn't baked with proper recipies / feature flag (https://github.com/Connexions/webview/pull/2138)


helenemccarron:
We could use text-indent and padding but I agree with you that flex is much easier and a better option.

Issue #2141, opened 2018-12-19
Small distance between element and Hide Solution


wilkus27:
On katalyst01 (e.g. https://katalyst01.cnx.org/contents/AndhM9Sd@4.13:kWoglDtR@4/8-8-Use-the-Complex-Number-System) different elements are too close option Hide Solution:

mathematical object:

![hide_math](https://user-images.githubusercontent.com/40228854/50217740-e0e4ba80-0389-11e9-96a5-d32b3f16fcdd.JPG)

paragraph:

![hide_par](https://user-images.githubusercontent.com/40228854/50217755-eb06b900-0389-11e9-81e3-8ee67af0e934.JPG)

wilkus27:
Verified on katalyst01:

1. 
![math](https://user-images.githubusercontent.com/40228854/51525699-2c7b0000-1e31-11e9-886d-1b02a49edb54.JPG)

2. 
![par](https://user-images.githubusercontent.com/40228854/51525720-3997ef00-1e31-11e9-8fd6-f3e1d25ebfa1.JPG)

Checked also in **Prealgebra** and **Elementary Algebra**.

PR #2145, opened 2018-12-20
added margin-top to os-solution-container paragraphs in examples

Adrianltd:
#2141

### DIFF
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -245,6 +245,14 @@ h1.example-title .text {
   }
 }
 
+[data-type="example"] {
+  .os-solution-container {
+    p {
+      margin-top: 2rem;
+    }
+  }
+}
+
 // Equations
 // --------------------------------------------------
 .equation,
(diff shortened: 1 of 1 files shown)


src/scripts/modules/media/body/content-baked.less
```diff
@@ -245,6 +245,14 @@ h1.example-title .text {
   }
 }
 
+[data-type="example"] {
+  .os-solution-container {
+    p {
+      margin-top: 2rem;
```
PiotrKozlowski: This will add margin to all `p` inside `.os-solution-container`.
```suggestion
      .os-solution-container { margin-top: 2rem; }
```

Please consider that this issue is affecting not only paragraphs:
screenshot_2018-12-20 intermediate algebra - openstax cnx

Adrianltd: I've added margin to `p` because there's also no gap between the solutions:
image
As far as I've checked, all solutions are inside a `p` tag or start with a `p` tag.  There are some that start with a table, but it looks fine.
So still we would need to add a margin to `p`, because of the few versions of solutions + eventually a margin to a whole `.os-solution.container`
`.os-solution-container {
    margin-top: 1rem;
    p {
      margin-top: 1rem;
    }
  }`
What do you think @PiotrKozlowski ?

PiotrKozlowski: I think that will be good solution

Issue #2121, opened 2018-12-07
Sim does not display on Webview


edwoodward:
Tom noticed a sim in an iframe never loads. The sim works fine outside of the iframe. The error in the console is...
```javascript
Uncaught TypeError: Cannot read property 'y' of undefined
    at L.n.getDynamicFieldAt (optimized.js:172)
    at L.n.evaluateLatticePoint (optimized.js:173)
    at L.n.update (optimized.js:173)
    at L.n.updateMVT (optimized.js:173)
    at L.n.initialize (optimized.js:173)
    at L.n.s (optimized.js:172)
    at new L.n (optimized.js:169)
    at L.n.initFieldLattice (optimized.js:173)
    at L.n.initGraphics (optimized.js:173)
    at L.n.postRender (optimized.js:172)
```
Content is at...
https://cnx.org/contents/Ax2o07Ul@14.4:R_XAX-aM@9/24-2-Production-of-Electromagnetic-Waves

The sim is at...
https://archive.cnx.org/specials/c8dd764c-ae74-11e5-af4c-3375261fa183/radio-waves/#sim-radio-waves

Bkblodget:
I looked into this as much as I could and this is what I've found:

1. as @edwoodward pointed out this sim works fine in archive. It also works fine in legacy: https://legacy.cnx.org/content/m42440/1.9/?collection=col11406/1.14&legacy=true

2. I found another CNX/archive sim in College Physics that works just fine & with the exact same markup, so I am ruling out cnxml markup as the issue: https://cnx.org/contents/Ax2o07Ul@14.4:YCO4fVoo@9/2-4-Acceleration

Maybe the cause is XHTML validation mentioned in https://github.com/Connexions/baked-pdf/issues/5 . Will require further investigation from devs.

Bkblodget:
This issue is similar to https://github.com/Connexions/webview/issues/2099 and updates will be tracked there, however, this issue will remain open until fix is confirmed.

omehes:
Verify in the latest staging deploy. Looks good (a separate issue created for sims with scrollbars and overlaps).

{"date": "2019-02-04 09:55:02 CST",
 "webview": "9e7f3093433af0351140f86624fd5356a1278f53 (v0.45.0)",
 "cnx-archive": "4.8.0",
 "cnx-publishing": "0.15.0",
 "oer.exports": "astonishing.applepie v0.71.0",
 "press": "8.0.2",
 "cnx-deploy": "v2.36.0"

PR #2153, opened 2019-01-08
Allow sims to use width specified in markup

edwoodward:
Width of 100% was overriding the width in the HTML. Removing it allows the sims to display correctly.


edwoodward:
Also fixes Sim does not display on Webview - #2121

PR #2161, opened 2019-01-14
Revert change for empty collections

edwoodward:
Change was causing an endless loop on the server


reedstrm:
well, on the client

PR #2162, opened 2019-01-15
issue-2086-2091

brittany-johnson:
addresses issues: 
https://github.com/openstax/webview/issues/2086
https://github.com/openstax/webview/issues/2091

### DIFF
--- a/src/scripts/modules/search/advanced/advanced.coffee
+++ b/src/scripts/modules/search/advanced/advanced.coffee
@@ -69,12 +69,10 @@ define (require) ->
 
       @search(query)
 
-    search: (query) ->
-      if !query
+    search: (query,values) ->
+      if !values || values == 'sort:popularity' || values == 'sort:pubDate'
         $("#alert-danger").show()
-      else if query == 'sort:popularity' || query == 'sort:pubDate'
-          $("#alert-danger").show()
-      else if query
+      else
         router.navigate("search?q=#{query}", {trigger: true})
 
     formatQuery: (obj) ->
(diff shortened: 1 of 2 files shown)


src/scripts/modules/search/advanced/advanced.coffee
```diff
@@ -69,11 +68,14 @@ define (require) ->
...
-        $("#alert-danger").show()
 
     search: (query) ->
-      router.navigate("search?q=#{query}", {trigger: true})
+      if !query
+        $("#alert-danger").show()
+      else if query == 'sort:popularity' || query == 'sort:pubDate'
+          $("#alert-danger").show()
+      else if query
+        router.navigate("search?q=#{query}", {trigger: true})
```
tomjw64: It might yield more stable behavior to run this validation on `values` rather than `query`, to avoid string comparison

brittany-johnson: @tomjw64 updated it! Can you re-review?

src/scripts/modules/search/advanced/advanced.coffee
```diff
@@ -69,11 +68,12 @@ define (require) ->
       query = @formatQuery(values)
 
       @search(query)
-      if !query
-        $("#alert-danger").show()
 
-    search: (query) ->
-      router.navigate("search?q=#{query}", {trigger: true})
+    search: (query,values) ->
```
tomjw64: Since validation is a separate action than a search, why don't we make a separate `validateSearchable` function for the query values. At the same time, we can take advantage of using the full data structure of `values` instead of `query` to validate. Hint: `values.filter(function(field) { return field.value !== "" }).every(function(field) { return field.name === "sort"})` might help you determine if a query is searchable or not

PR #2178, opened 2019-01-30
issue-2160

brittany-johnson:
https://github.com/openstax/webview/issues/2160


src/scripts/modules/media/body/content-baked.less
```diff
@@ -250,6 +250,17 @@ h1.example-title .text {
   }
 }
 
+[data-type="exercise"] {
+  [data-type="problem"] {
+    p {
```
tomjw64: Should we add something to specify that only the first paragraph follows the number inline?

brittany-johnson: Good idea, just made another commit @tomjw64

PR #2208, opened 2019-03-01
removed dots from unordered lists in the preface

Adrianltd:
https://github.com/openstax/webview/issues/2198

### DIFF
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -482,6 +482,10 @@ h1.example-title .text {
   display: inline;
 }
 
+.preface ul {
+  list-style-type: none;
+}
+
 .appendix [data-type="list"] {
   margin-top: 1rem;
 }


Adrianltd:
@openstaxalina I have removed those dots in webview, but @wilkus27 just noticed me that previously this was removed from the content https://github.com/openstax/content-managers/issues/500
Should my PR be removed and content managers will remove those dots?

openstaxalina:
@Adrianltd thank you for pointing this out.  the reason we decided to remove the dots via the content in other titles is because @helenemccarron notified us (see comments on 
openstax/content-managers#500) that the change would impact all lists in recipes.  

What content is impacted by this PR?

Adrianltd:
@openstaxalina only unordered lists inside the `Preface`, so it affects dotted lists only in the `Preface`.

openstaxalina:
@Adrianltd great, thank you for letting me know.  This is an acceptable change.  Thanks again!

Issue #2203, opened 2019-02-28
Solution in Example isn't bold


MartaSztuka:
In each chapter's subsection 'Exercise' should have bolded 'Solution'
https://katalyst01.cnx.org/contents/i4nRcikn@3.7:DSFrV7ZX@4/6-1-Areas-between-Curves

PDF:
![screenshot from 2019-02-28 13-35-06](https://user-images.githubusercontent.com/47241114/53567157-9ee0ad80-3b5e-11e9-8b59-694c7b865383.png)

webview:
![screenshot from 2019-02-28 13-34-51](https://user-images.githubusercontent.com/47241114/53567173-aef88d00-3b5e-11e9-8a8b-1116ca3d1484.png)

This should be bold

openstaxalina:
Can these be targeted using `data-type="solution-title"`?
![Screen Shot 2019-03-25 at 1 25 08 PM](https://user-images.githubusercontent.com/7863733/54944456-82d3f000-4f01-11e9-89ac-288cffc33004.png)

openstaxalina:
If markup is inconsistent, we need to standardize the markup for the Solution header in Examples.

wilkus27:
Will be resolved in recipe. I move it to this repo.

wilkus27:
Issue moved to [openstax/cnx-recipes #975](https://github.com/openstax/cnx-recipes/issues/975) via [**ZenHub**](https://www.zenhub.com/)

PR #2209, opened 2019-03-05
Bolded solution title in Example #2203

wilkus27:
Title Solution in Example should be bolded - issue https://github.com/openstax/webview/issues/2203


wilkus27:
This problem is resolved in recipe: https://github.com/openstax/cnx-recipes/pull/965.

PR #2223, opened 2019-03-13
issue-2196

brittany-johnson:
https://github.com/openstax/webview/issues/2196

### DIFF
--- a/src/scripts/modules/media/header/header.coffee
+++ b/src/scripts/modules/media/header/header.coffee
@@ -87,6 +87,9 @@ define (require) ->
       @listenTo(@model, 'change:currentPage.searchTitle', @render)
 
     onRender: () ->
+      #allows user to access 'skiptocontent' and header navigation when navigating from pg to pg in a book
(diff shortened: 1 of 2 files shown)


src/scripts/modules/media/header/header.coffee
```diff
@@ -87,6 +87,9 @@ define (require) ->
       @listenTo(@model, 'change:currentPage.searchTitle', @render)
 
     onRender: () ->
+      #allows user to access 'skiptocontent' and header navigation when navigating from pg to pg in a book
+      $(':root').attr("tabindex",0).focus()
```
tomjw64: Do you have to dynamically set the tabindex for the root here and not in the page template itself? (I legit don't know)

brittany-johnson: @tomjw64 I just tried to place it on the template and it technically works but when a new page loads outside of the book view, you have to tab twice to reach the first viewable element. 
I don't know why this is

PR #2231, opened 2019-03-29
Removed bold from links inside sections

wilkus27:
Issue: https://github.com/openstax/webview/issues/2215.

### DIFF
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -259,14 +259,23 @@ h1.example-title .text {
   }
 }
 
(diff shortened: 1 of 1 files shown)


wilkus27:
This problem is a regression from solution of issue https://github.com/openstax/webview/issues/2160.

Now links are bolded only on composite pages.

However I had to copy some part of the code:
```
[data-type="exercise"] {
  [data-type="problem"] {
    p:first-of-type {
      display: inline-block;
    }
  }
}
```
because it is resolving problem https://github.com/openstax/webview/issues/2201 which occurs on regular page not composite page.

PR #2238, opened 2019-04-03
changed figcaption display to block so unnumbered captions display correct

Adrianltd:
This should solve https://github.com/openstax/webview/issues/2237 and https://github.com/openstax/webview/issues/2242

### DIFF
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -106,6 +106,12 @@ iframe {
   }
 }
 
+figure:not([data-orient="vertical"]) {
+  > figcaption {
+    display: block;
+  }
+}
+
 .os-figure {
   display: table;
   margin: 3rem auto;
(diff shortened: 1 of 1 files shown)


src/scripts/modules/media/body/content-baked.less
```diff
@@ -106,6 +106,15 @@ iframe {
   }
 }
 
+figure.splash.unnumbered {
+  display: table;
```
tomjw64: Can this change be made using a `block` or `inline block` display or by using flexbox with a column flow? Typically `display: table` is no longer required for most tasks.

Adrianltd: @tomjw64 That's a copy-paste from the styling for `os-figure` (see line 122 here https://github.com/openstax/webview/blob/0a1d53c49152eb82da6d6d6788886275b687c96d/src/scripts/modules/media/body/content-baked.less ). I thought that it was set like this for a reason which I can't test and didn't want to do the very same styling in a different way. 
Should I update the styling for both `figure.splash.unnumbered` and `os-figure`?

tomjw64: Fair point. Approved, in that case.

Issue #2228, opened 2019-03-28
Line break in exercises


uskka:
There is a line break after the exercise number on webview but it only appears on Safari and in exercises that start with equation and there is a figure after it. 

https://katalyst01.cnx.org/contents/HTmjSAcf@3.4:DSFrV7ZX@5/2-1-Areas-between-Curves
https://katalyst01.cnx.org/contents/oxzXkyFi@3.7:9fEOYsBF@8/3-3-Arc-Length-and-Curvature

Webview
<img width="909" alt="Screenshot 2019-03-28 at 10 12 12" src="https://user-images.githubusercontent.com/40228918/55148229-1bf44800-5148-11e9-8fca-e9bad85b0b4c.png">

PDF
<img width="1309" alt="Screenshot 2019-03-28 at 10 12 20" src="https://user-images.githubusercontent.com/40228918/55148230-1bf44800-5148-11e9-97ff-24c35438a7e7.png">

wilkus27:
After discussion in https://github.com/openstax/webview/pull/2239#issuecomment-482634139 decision is to not fix this issue.

PR #2239, opened 2019-04-04
#2228 line break in exercises

wilkus27:
Issue: https://github.com/openstax/webview/issues/2228

Removed line break in section exercises when paragraph in `os-problem-container` has only mathjax element. Problem occured only on Safari.

### DIFF
--- a/src/scripts/modules/media/body/content-baked.less
+++ b/src/scripts/modules/media/body/content-baked.less
@@ -253,7 +253,10 @@ h1.example-title .text {
   .os-solution-container {
     display: inline;
     > :first-child:not(ul):not(ol):not([data-type="note"]):not(.os-figure) {
-      display: inline-block;
+      display: inline;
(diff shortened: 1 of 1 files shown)


wilkus27:
@tomjw64 I tried something like that. I wasn't sure if it thats what you mean. But it doesn't work. I had to commit this change because I don't have a Safari, so it was one way to test it. The previous change was working properly.

tomjw64:
Unfortunately, I am worried that setting `inline-block` will cause unwanted regressions. For example, when using `inline-block`, I get unwanted line-breaks on exercises 48-57 here: https://katalyst01.cnx.org/contents/HTmjSAcf@3.4:DSFrV7ZX@5/2-1-Areas-between-Curves

wilkus27:
Ok, thank you @tomjw64 for your investigation. I will close this PR and add comment to the issue.

PR #2250, opened 2019-05-14
Upgrade from Node 6.9.x to 10.15.x

pumazi:
This moves from our custom container to using the official node built container. Last time I tried upgrading it failed, so something has changed between now and then that has allowed it to build.

I'm not exactly sure how to test this; so please use scrutiny while reviewing.


pumazi:
I've assigned to @helenemccarron for merge acceptance of this work.

helenemccarron:
I second Thomas, it looks good but Webview should be tested to ensure nothing odd is happening after the update.

PR #2253, opened 2019-05-22
changed default table styles

Adrianltd:
This PR will solve:
https://github.com/openstax/webview/issues/2154
https://github.com/openstax/webview/issues/2183
https://github.com/openstax/webview/issues/2184

It will also affect any other content that is created in tables eg `tables with equations`. 
https://katalyst01.cnx.org/contents/mwjClAV_@8.2:SXjiAlpU@8.2/Key-Equations

@openstaxalina This should look good also for tables with merged cells, paddings in cells also look good. It will be verified by QA.
I was thinking if we should make the `table head` border darker or thicker so it can be distinguished from the table body?

### DIFF
--- a/src/scripts/modules/media/body/content-common.less
+++ b/src/scripts/modules/media/body/content-common.less
@@ -422,13 +422,17 @@ table {
     tr > td[data-align="right"]{
       text-align: right;
     }
+    tr > td {
+      padding: 1em;
+      border: 0.1rem solid @table-border-color;
+    }
   }
   tfoot {
     > tr {
       > th,
       > td {
         padding: @table-cell-padding;
-        border-top: 0.1rem solid @table-border-color;
+        border: 0.1rem solid @table-border-color;
         line-height: @line-height-base;
         vertical-align: top;
       }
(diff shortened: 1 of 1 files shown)


lwtchu:
Can someone provide a screenshot of what it looks like now so I can give feedback?

Ideally - yes the header should have a thicker or darker line. 

Screen Shot 2019-05-28 at 3 07 14 PM
Currently there's a 1px(?) line for the header. The body table borders can be 0.5px if that makes it easier.

helenemccarron:
@wilkus27 I know this wasn't your PR but would you like to take over?

wilkus27:
@helenemccarron of course.
So I checked locally how it looks: 

image

For comparison this is how this table looks currently on katalyst01: (https://katalyst01.cnx.org/contents/kg0cimBs@14.1:Wqg4xZyG@7/6-3-Calculate-Activity-Based-Product-Costs)

image

Do we want to do any changes?

openstaxalina:
Thank you @wilkus27!  We will provide feedback today.

lwtchu:
@wilkus27 looks BEAUtiFUL. can you reduce the padding on the inside by half?
Screen Shot 2019-05-31 at 9 30 40 AM

wilkus27:
@lwtchu Do you mean only in header or in every cell?

Here padding is reduced in every cell and from each side (left/right top/bottom):

image

Here, in every cell but only from top and bottom, left and right padding is the same as previously:

image

What looks better?

wilkus27:
@lwtchu @openstaxalina What do you think about tables in examples, where background is grey?

With all of those changes it looks like that:
image

Currently on katalyst01: (https://katalyst01.cnx.org/contents/yqV9q0HH@18.1:SlIIi4zi@21/1-3-Subtract-Whole-Numbers)

image

openstaxalina:
@wilkus27 thank you!  Larissa is out this week, so I will respond:

Please implement the second option for padding reduction, in which you reduced padding in every cell, but only top/bottom.

For the tables inside notes, please implement it as shown in your first screenshot ("With all of those changes it looks like that:").  

Thank you again!

PR #2255, opened 2019-05-30
Do a hard request for featured book clicks

pumazi:
We are setting the `data-bypass` attribute on the anchor to allow the default
behavior to proceed. This passes the requested link to the browser rather
than having the javascript capture it for internal requesting of new content.

We make the browser do a request of these URLs so that it contacts
the webserver (nginx) to do redirects of REX books (aka all openstax books).

This is associated with openstax/cnx#401


pumazi:
@brittweinstein , I'm unable to attach this to openstax/cnx#401. I'd guess it's because webview is in some other ZenHub group or something of that nature?

brittweinstein:
@pumazi this is odd! I'm not sure what is causing that problem. In any case, I'll move this to the In Progress/Code Review pipeline so that we can keep eyes on it. (I'll also put  the milestone info)

BryanEliDimas:
@pumazi Yeah, webview would need to be added to the CNX Zenhub workspace in order to attach this to an issue there. cc: @brittweinstein

PR #2272, opened 2019-09-06
Removed link to Android app

edwoodward:
The Android app was removed from the Google Play Store so the link in the footer no longer works.

### DIFF
--- a/src/scripts/modules/footer/footer-template.html
+++ b/src/scripts/modules/footer/footer-template.html
@@ -20,13 +20,6 @@
     </div>
   </section>
 
-  <section class="connect">
-    <h1 class="small-header" data-l10n-id="all-footer-connect">Connect</h1>
-    <ul>
-      <li><a href="https://play.google.com/store/apps/details?id=org.openstaxcollege.android" data-l10n-id="all-footer-android-app">Android App</a></li>
-    </ul>
-  </section>
-
   <section class="share">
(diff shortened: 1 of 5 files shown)


src/scripts/modules/footer/footer-template.html
```diff
@@ -20,13 +20,6 @@
     </div>
   </section>
 
-  <section class="connect">
```
brenguyen711: Jello! Should we remove all the code for `class="connect"` and `data-l10n-id="all-footer-connect"` as well? It looks like it's only used here.

karenc: Yeah I think it's safe to remove it.  I did a search on the webview code:

```
karen@yachiyo:~/src/webview$ git grep '\<connect\>'
src/locale/en-US/dictionary.ftl:all-footer-connect = Connect
src/locale/pl/dictionary.ftl:all-footer-connect = Łącza
src/scripts/modules/footer/footer.coffee:      @$el.find('.connect').height(height)
src/scripts/modules/footer/footer.less:  > .connect, > .share {
src/scripts/modules/footer/footer.less:  > .connect {
```

Looks to me they can all be removed.

brenguyen711:
before:
Screen Shot 2019-09-13 at 2 36 00 PM

after:
Screen Shot 2019-09-13 at 2 35 44 PM

PR #2284, opened 2019-11-12
Make webview UI run inside docker-compose with other cnx apps

tomjw64:


brittany-johnson:
@tomjw64 What do you recommend I do to test your changes? Will running `./script/start` be sufficient?

.dockerfiles/docker-entrypoint.sh
```diff
@@ -41,13 +41,15 @@ if [[ "$1" = 'nginx' ]]; then
...
-	file_env 'ARCHIVE_HOST'   "${ARCHIVE_HOST:-archive.cnx.org}"
-	file_env 'ARCHIVE_PORT'   "${ARCHIVE_PORT:-80}"
-	file_env 'OPENSTAX_HOST'  "${OPENSTAX_HOST:-openstax.org}"
-	file_env 'OPENSTAX_PORT'  "${OPENSTAX_PORT:-443}"
-	file_env 'EXERCISES_HOST' "${EXERCISES_HOST:-exercises.openstax.org}"
-	file_env 'EXERCISES_PORT' "${EXERCISES_PORT:-443}"
-	file_env 'LEGACY_HOST'    "${LEGACY_HOST:-legacy.cnx.org}"
+	file_env 'ARCHIVE_HOST'     "${ARCHIVE_HOST:-archive.cnx.org}"
+	file_env 'ARCHIVE_PORT'     "${ARCHIVE_PORT:-80}"
+	file_env 'FE_ARCHIVE_HOST'  "${FE_ARCHIVE_HOST:-archive.cnx.org}"
```
karenc: RE: commit message

> An example:
>     FE_ARCHIVE=archive-local.cnx.org  # maybe in /etc/hosts
>     FE_ARCHIVE_PORT=80
>     ARCHIVE=archive  # recognized as host in the docker-compose network
>     ARCHIVE_PORT=6543

Why don't you just use `ARCHIVE=archive-local.cnx.org`?

tomjw64: The `ARCHIVE` env variable is used in the webview container's nginx config for a proxy pass, so the host must be recognized by the container. An entry `archive-local.cnx.org` in the `/etc/hosts` of the docker host is not recognized from inside the container , so the webview container would fail and exit. The reason I leave `archive.cnx.org` as the default in both cases is that, running in isolation, a webview container's nginx and its frontend client would both recognize it.

tomjw64:
@brittany-johnson this only modifies docker stuff, so whatever you decide to try, it should be via starting webview with docker-compose.

.dockerfiles/docker-entrypoint.sh
karenc: Ah so basically because it's not set up using real dns, we have this problem...

Do you know if there's any documentation about how to use webview in docker so you can add this new bit to it?

tomjw64: There's a tiny section in the README. I'll tack a little bit of information on there, since this use case is very specific to it. :)

tomjw64: @karenc I have added notes to the README

